Issue #2, opened 2019-11-22
Publishing Helm charts via packages project


thjaeckle:
Yesterday we had a discussion with a potentially new Ditto adopter @axdotl from kiwigrid on Gitter about changes to the Ditto Helm chart which he already contributed as PR.

We also discussed about the packaging and hosting of our Helm chart. The Ditto project doesn't host it yet, so kiwigrid did that, see: https://hub.helm.sh/charts/kiwigrid/ditto-digital-twins

I don't know whether e.g. Hono does host its Helm chart (packaged and available via a website and `index.yaml`), but I would like to host the "official" Helm charts of our projects by ourself so that improvements to the chart are added in a single place.

That's btw how kiwigrid's index.yaml looks like: https://kiwigrid.github.io/index.yaml

I think it would be a great idea to use the Eclipse IoT Packages project to distribute various Helm charts used in the packages, e.g. via the project site as a subfolder https://eclipse.org/packages/charts/ or via a github.io site of this repo (like kiwigrid does it).

That way we could simply document to add a Helm repo:
```
$ helm repo add eclipse-iot https://eclipse.org/packages/charts/
```

And afterwards all `eclipse-iot` charts could automatically be installed, e.g.:
```
$ helm install eclipse-iot/hono
$ helm install eclipse-iot/ditto
```

Or, for the packages:
```
$ helm install eclipse-iot/package-zero
```

Any ideas/comments on that?

ctron:
I absolutely like that idea!

thjaeckle:
What do you think about where to manage the Helm chart sources?

I see two options:
* Either copy the Helm sources from the project's GitHub repos to this repo and keep them manually in sync
* Have this repo contain the "only truth" for Eclipse IoT Packages Helm charts, delete the Helm chart sources from the project's Github repos

axdotl:
I would prefer the 'single-point-of-truth' option. We would be glad, to support on this!

I suggest to do it in a similar way we do it for the kiwigrid/helm-charts. This means to have all charts in a single place (e.g. a chart-directory in `eclipse/packages/`), verify and package the charts with CircleCI and publish them via Github pages (maybe `eclipse.github.io`).
Afterwards add the created registry to Helm Hub: https://github.com/helm/hub to have the charts easy accessible for users.

thjaeckle:
Using `eclipse.github.io/packages/` would work I think with the "GH pages" approach.
[...]

PR #3, opened 2019-11-26
contribution of Eclipse Ditto helm chart

thjaeckle:
Based on the Helm rework contribution of Kiwigrid https://github.com/eclipse/ditto/pull/553 
Using a new folder structure as mentioned in #2 

Still to verify if that is working as expected - a review from @axdotl wold be much appreciated :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,12 @@
+# IDE specific ignores (IntelliJ, Eclipse, VScode):
 .idea
 *.iml
 .project
 .settings
 .metadata
 .vscode/
-target/
-*.mdb
-.env
-.DS_Store
+# Helm and kind specific ignores:
 Chart.lock
 kind-config.yaml
+# dependencies of charts get downloaded to chart/<chartname>/carts - so exclude those:
 charts/*/charts
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,13 @@
+.idea
+*.iml
+.project
+.settings
+.metadata
+.vscode/
+target/
+*.mdb
+.env
+.DS_Store
+Chart.lock
+kind-config.yaml
+charts/*/charts
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -31,6 +31,7 @@ The project maintains the following source code repositories:
 
 # Copyright holders
 
+* Copyright 2019 Red Hat Inc
 * Copyright 2019 Bosch Software Innovations GmbH
 * Copyright 2019 Kiwigrid GmbH
 
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -29,6 +29,11 @@ The project maintains the following source code repositories:
 
 * https://github.com/eclipse/packages
 
+# Copyright holders
+
+* Copyright 2019 Bosch Software Innovations GmbH
+* Copyright 2019 Kiwigrid GmbH
+
 ## Third-party Content
 
 This project leverages the following third party content.
--- a/charts/ditto/README.md
+++ b/charts/ditto/README.md
@@ -1,0 +1,72 @@
+# Eclipse Ditto
+
+## Introduction
+
+[Eclipse Ditto™](https://www.eclipse.org/ditto/) is a technology in the IoT implementing a software pattern called “digital twins”.
+A digital twin is a virtual, cloud based, representation of his real world counterpart (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations, …).
+
+This chart uses `eclipse/ditto-XXX` containers to run Ditto inside Kubernetes.
+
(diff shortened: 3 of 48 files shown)


axdotl:
I'll have a look on it later this week. Thanks for your patience!

ctron:
This looks great. I would like to understand one thing before taking a deeper look into this: why is the original author of this? Is this you @thjaeckle?

thjaeckle:
No, that's not so easy to define :D
The main author of the "legacy Helm chart" was @stmaute together with a colleague @ghandim
then there were some adjustments by Microsoft
and the main author of the rewrite was @axdotl 
I just adjusted some parts of the chart before copying it here (e.g. not copying the openAPI files, but using them from GitHub).

ctron:
Ok, then I guess we need a proper CQ for that :grin: … could you get this started?!

thjaeckle:
A CQ for a Helm chart which is super special for Eclipse Ditto, really? :D

charts/ditto/templates/_helpers.tpl
```diff
@@ -0,0 +1,55 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "eclipse-ditto.name" -}}
```
axdotl: Usually the properties are named as the chart, means `ditto` i.s.o. `eclipse-ditto`.

ctron:
> A CQ for a Helm chart which is super special for Eclipse Ditto, really? :D

Yes. It is not a dependency CQ, but project code CQ. It is more than 1000 lines of code, and the source is worth investigating. So I think we do need a project code CQ.

charts/ditto/README.md
```diff
@@ -0,0 +1,72 @@
...
+## Prerequisites
+
+* Has been tested on Kubernetes 1.11+
+
+## Installing the Chart
+
+To install the chart with the release name `eclipse-ditto`, run the following command:
+
+```bash
+helm install kiwigrid/eclipse-ditto --name eclipse-ditto
```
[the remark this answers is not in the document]
thjaeckle: Yes, I'm however not yet sure which final name we will use for the "repo"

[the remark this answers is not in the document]
ctron: Let's discuss that in a separate issue.

thjaeckle:
CQ was approved:

> --- Comment #12 from Shawn Kilpatrick <shawn.kilpatrick@eclipse-foundation.org>  2019-12-04 08:22:03 --- We have completed a high level triage of the attachment based on Parallel IP [1].  You may proceed to check the content into the Eclipse repository.  We will perform a comprehensive analysis at a later date based on IP work queue.
[1] http://www.eclipse.org/projects/handbook/#ip-parallel-ip

.gitignore
```diff
@@ -0,0 +1,13 @@
+.idea
```
ctron: I would prefer to have a list which is more tailored towards the content of this repository. And maybe add some comments of "why". At least on the top level of the repository.

NOTICE.md
```diff
@@ -29,6 +29,11 @@ The project maintains the following source code repositories:
 
 * https://github.com/eclipse/packages
 
+# Copyright holders
+
+* Copyright 2019 Bosch Software Innovations GmbH
```
ctron: As we are adding this section, I would like to see "Red Hat Inc" there as well :wink:

.gitignore
thjaeckle: So should I comment each `.gitignore` entry?

thjaeckle: done ..

NOTICE.md
thjaeckle: done ..

ctron:
Ok. So my understanding is, that this PR only adds the Ditto Helm chart. But doesn't put this into the "Package Zero"?! If that is the case, then how would be include this into the deployment of "Package Zero"?

thjaeckle:
Correct, IMHO we should adjust the folder structure like discussed in #2 :
- put all charts in `charts` folder
- packages in `packages`
- make the Helm charts in `packages` umbrella packages, like e.g. Shown here: https://github.com/CodeJjang/helm-microservices-example

That way charts may be used in several packages.
I didn't want to do that before discussing and the discussion in #2 never started about the folder structure.

We could do that in a follow up PR moving the Hono package (while also adding copyright headers) and establishing the umbrella package zero chart.

ctron:
> We could do that in a follow up PR moving the Hono package (while also adding copyright headers) and establishing the umbrella package zero chart.

Sounds like a plan. We also need to update the Hono charts to more recent ones.

Ok, so the plan is to merge this and then change the structure? That would be ok for me.

Issue #5, opened 2019-12-02
Create a Github actions Helm lint / install pipeline


monotek:
As decided in the call we had (https://github.com/eclipse/packages/issues/4) we'll create a Github actions pipeline, which will lint (via https://github.com/helm/chart-testing) and install the Eclipse Helm charts.

We'll use Kubernetes Kind (https://github.com/kubernetes-sigs/kind) to install the charts in the last 3 recent Kubernetes versions.

Publishing the charts will be done in a later task. Likely this will be tracked in the following Ticket:
https://github.com/eclipse/packages/issues/2

ctron:
I just saw that GitHub actions now also allows to hook up customer provided compute resources. Maybe that would be an alternative if we hit some kind of resource limitation.

monotek:
We managed to switch our pipeline (https://github.com/kiwigrid/helm-charts) from Circle CI to Github actions now but still have to fix some workarounds, which had to be introduced. 

But i guess we should be ready to go some time next week :)

monotek:
We moved from our own bash scripts to the official Helm Github actions but using some additions.

We've created some PRs for the actions which speed up ci a bit:
* https://github.com/helm/chart-testing-action/pull/11
* https://github.com/helm/kind-action/pull/6

Current state of the pipeline can be seen in this PR:
* https://github.com/kiwigrid/helm-charts/pull/252

So we're ready to use this. If the Github Actions PRs are not merged we're going to use our modified version or just stick with the original version, which would be just a bit slower.

thjaeckle:
No, we don't have a deadline.

Great to see the progress you're making.

monotek:
I've create a PR: https://github.com/eclipse/packages/pull/9

monotek:
closed by: https://github.com/eclipse/packages/pull/9

thjaeckle:
Verified by test PR #10

PR #9, opened 2019-12-16
added github action helm ci pipeline

monotek:

As discussed in https://github.com/eclipse/packages/issues/5 this is the pr to implement an ci pipeline for the Helm charts via Github actions.

Currently we have more than 3 K8s versions enabled, also because the last 3 would not match our current environment, as 1.14 is latest available version on GKE.

The extra job check-for-chart-changes is just a workaround to speedup ci and to not waste resources, if there are no chart changes.


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,57 @@
...
+          - v1.15.3
+          - v1.16.2
+          - v1.17.0
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Create kind cluster
+        uses: helm/kind-action@master
+        with:
+          config: /home/runner/work/helm-charts/helm-charts/.ci/kind-config.yaml
```
thjaeckle: `kind-config.yml` is no part of this PR, is it?

.ci/check-for-chart-changes.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
thjaeckle: could we put the shell files, template files, etc. for CI in the `.github` folder as well?
using another directory `.ci` artificially splits the CI stuff.

.ci/check-for-chart-changes.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# check for chart changes to speedup ci
+#
+
+set -o errexit
+set -o pipefail
+
+CHART_REPO="https://github.com/kiwigrid/helm-charts.git"
```
thjaeckle: That should be replaced with the `https://github.com/eclipse/packages.git`, correct?

.ci/ct.yaml
```diff
@@ -0,0 +1,5 @@
+helm-extra-args: --timeout 7200
```
thjaeckle: could you please add a license header to this file?
```
# Copyright (c) 2019 Contributors to the Eclipse Foundation
#
# See the NOTICE file(s) distributed with this work for additional
# information regarding copyright ownership.
#
# This program and the accompanying materials are made available under the
# terms of the Eclipse Public License 2.0 which is available at
# http://www.eclipse.org/legal/epl-2.0
#
---
```

.ci/check-for-chart-changes.sh
thjaeckle: Also: a license header for all shell files would be needed (below the shebang I guess):
```
# Copyright (c) 2019 Contributors to the Eclipse Foundation
#
# See the NOTICE file(s) distributed with this work for additional
# information regarding copyright ownership.
#
# This program and the accompanying materials are made available under the
# terms of the Eclipse Public License 2.0 which is available at
# http://www.eclipse.org/legal/epl-2.0
#
```

.ci/lint-scripts.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+#
+# lint bash scripts
+#
+
+set -o errexit
+
+CONFIG_DIR=".ci"
```
thjaeckle: I would suggest using `.github/config`

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Lint Bash scripts
+        uses: docker://koalaman/shellcheck-alpine:v0.7.0
+        with:
+          args: /github/workspace/.ci/lint-scripts.sh
+      - name: Check for chart changes
+        run: /home/runner/work/helm-charts/helm-charts/.ci/check-for-chart-changes.sh
```
thjaeckle: this path depends on the GH repository name (`helm-charts/helm-charts`), correct?
So that should be adjusted to the "packages" repo.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    needs: check-for-chart-changes
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Run chart-testing (lint)
+        uses: helm/chart-testing-action@master
+        with:
+          command: lint
+          config: /home/runner/work/helm-charts/helm-charts/.ci/ct.yaml
```
thjaeckle: Again, path has to be adjusted

.ci/check-for-chart-changes.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo "Check for chart changes to speedup ci..."
+
+git remote add chart-changes "${CHART_REPO}"
+git fetch chart-changes master
+
+if [ -z "$(git diff --find-renames --name-only "$(git rev-parse --abbrev-ref HEAD)" remotes/chart-changes/master -- charts)" ]; then
```
thjaeckle: Couldn't the same (what the complete script does) be achieved with a GitHub Action containing the `paths` configuration?
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths

.ci/check-for-chart-changes.sh
ctron: I would opt for `#!/usr/bin/env bash`: https://stackoverflow.com/a/21613044/222044

.ci/check-for-chart-changes.sh
monotek: yes, but in this cae you would'nt get an error message but your workflow would just not start. imho this would be confusing for the user.

.ci/check-for-chart-changes.sh
monotek: done

.ci/check-for-chart-changes.sh
monotek: fixed

.ci/ct.yaml
monotek: added

.ci/lint-scripts.sh
monotek: done

.github/workflows/ci.yaml
monotek: replaced with apckages/packages

.github/workflows/ci.yaml
monotek: replaced with apckages/packages

.ci/check-for-chart-changes.sh
monotek: done

.ci/check-for-chart-changes.sh
thjaeckle: Ok, understood

.github/workflows/ci.yaml
thjaeckle: Could you add the missing `kind-config.yaml` (and also add the license header to it?) ? :)

monotek: also yaml files need the license header?

monotek: kind-config.yaml was missing because of: https://github.com/eclipse/packages/blob/master/.gitignore#L10

thjaeckle: ah, damn .. thanks - will adjust the .gitignore

.github/lint-scripts.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0
+#
+# SPDX-License-Identifier: EPL-2.0
+
+set -o errexit
+
+CONFIG_DIR=".ci"
```
thjaeckle: should be `CONFIG_DIR=".github"` now as the scripts were moved

monotek: ah thanks. missed that one :)

PR #12, opened 2019-12-18
Allow creating an OpenShift route

ctron:
This change adds support for deploying an OpenShift route, when setting `routes.enabled= true` (which is false by default) .


thjaeckle:
@ctron Chart needs a version bump :)

PR #14, opened 2020-01-09
Add info about chart publishing

axdotl:
Signed-off-by: Axel Köhler <axel.koehler@kiwigrid.com>

### DIFF
--- a/homepage/repository.md
+++ b/homepage/repository.md
@@ -18,7 +18,7 @@ are labeled explicitly.
 You can add the repository with a simple command:
 
 {% clipboard %}
-    helm repo add eclipse-iot https://eclipse.org/packages/charts
+    helm repo add eclipse-iot <https://eclipse.org/packages/charts>
 {% endclipboard %}
 
 This will add the repository, using the name `eclipse-iot`. Of course you may choose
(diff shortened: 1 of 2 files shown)


monotek:
pipline should'nt fail anymore after https://github.com/eclipse/packages/pull/17 is merged.

homepage/repository.md
```diff
@@ -18,7 +18,7 @@ are labeled explicitly.
 You can add the repository with a simple command:
 
 {% clipboard %}
-    helm repo add eclipse-iot https://eclipse.org/packages/charts
+    helm repo add eclipse-iot <https://eclipse.org/packages/charts>
```
thjaeckle: Hm, does the `<...>` mean that this one is optional?
Or will this command succeed with this syntax?

axdotl: My markdown linter mentioned this (see https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md034---bare-url-used)

If this might break something on the website, I'll revert it.

monotek:
@axdotl merge master to fix the ci job error.

homepage/repository.md
thjaeckle: As we don't need this to be a clickable link, I would remove the `<..>` again ..

Issue #16, opened 2020-01-09
Improve Helm chart CI pipeline


monotek:
At the moment we see a timeout for Hono. 
See: 
https://github.com/eclipse/packages/pull/13
https://github.com/eclipse/packages/pull/13/checks?check_run_id=356689661

Pipeline also fails when non chart changes are made.
See: 
https://github.com/eclipse/packages/pull/14
https://github.com/eclipse/packages/commit/0d5617c45244adcdec2aaa707d796b152648cec2/checks?check_suite_id=391476074

I'll try to improve this.

PR #17, opened 2020-01-09
improve helm ci pipline

monotek:

* only run when changes in charts dir
* fix timeout issue (configured timeout was not used as relative path for ct config file is needed)
* put lint bash scripts in own job
* updated kubernetes node versions to latest kind images

This will fix https://github.com/eclipse/packages/issues/16


monotek:
yes, changed it :)

PR #19, opened 2020-01-15
Reduce timeout for helm install command

sophokles73:
The original timeout of 7200 caused the installation to fail after 2
hours. Installation of charts should never take that long.
6 minutes seems to be a reasonable amount of time after which
installation can be considered to have failed.

### DIFF
--- a/.github/ct.yaml
+++ b/.github/ct.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2019 Contributors to the Eclipse Foundation
+# Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
 #
 # See the NOTICE file(s) distributed with this work for additional
 # information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


.github/ct.yaml
```diff
@@ -9,7 +9,8 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 
-helm-extra-args: --timeout 7200
+# consider helm install to be failed after 6 minutes
+helm-extra-args: --timeout 360
```
ctron: I think giving this a few minutes more is ok … something around 15 minutes sounds reasonable to me.

sophokles73: We are talking about installing a single chart here, aren't we? I wonder what chart would need that long to install. Can we meet in the middle, e.g. 10 minutes? We can increase once we have our first chart that doesn't complete installation in time ...

PR #20, opened 2020-01-15
Add requirements for charts

sophokles73:
It seems feasible to define some constraints and requirements regarding
the quality of the charts being hosted in the IoT Packages project.
This includes documentation, functionality, compliance etc.

The requirements have been copied (and slightly adapted) from the
central Helm Chart repo's CONTRIBUTING guide on GitHub.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,7 @@ Please make sure any file you newly create contains a proper license header like
 
 ````
 /**
- * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -62,7 +62,7 @@ You should, of course, adapt this header to use the specific mechanism for comme
 
 ````
 <!--
- Copyright (c) 2019 Contributors to the Eclipse Foundation
+ Copyright (c) 2020 Contributors to the Eclipse Foundation
 
  See the NOTICE file(s) distributed with this work for additional
  information regarding copyright ownership.
--- a/homepage/about.html
+++ b/homepage/about.html
@@ -25,7 +25,7 @@ is possible with Eclipse IoT, and enables you to try it out first hand.
 
 <p>
 IoT Packages are a combination of two or more Eclipse IoT projects, combined in an integrated way,
-leveraging the benefits of each project, to showcasing the benefit of their integration.
+leveraging the benefits of each project, to showcase the benefits of their integration.
 </p>
 
 <p>
@@ -34,7 +34,7 @@ And also enable them to try it out themselves, within minutes and on their own i
 </p>
 
 <p>
-Packages not only to provide deployments scripts, but also some description what the benefit of the
+Packages not only provide deployment scripts, but also some description what the benefit of the
 integrated package is, and some initial, tutorial like steps, to play with the setup, once it is deployed.
 </p>
 
@@ -60,8 +60,9 @@ base for all IoT packages.
 </p>
 
 <p>
-Helm is an tool which integrates nicely with Kubernetes. It is easy to use, and has the ability to create
-a set of YAML files, which are easy to deploy.
+Helm is a command line tool for managing Kubernetes applications. Helm Charts help define, install
+and upgrade even the most complex Kubernetes application.
+Packages consist of one or more Helm Charts which can be used to install the package to a Kubernetes cluster.
 </p>
 
 </div>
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -76,3 +76,61 @@ You should, of course, adapt this header to use the specific mechanism for comme
...
+* Follows [best practices](https://github.com/helm/helm/tree/master/docs/chart_best_practices)
+  (especially for [labels](https://github.com/helm/helm/blob/master/docs/chart_best_practices/labels.md)
+  and [values](https://github.com/helm/helm/blob/master/docs/chart_best_practices/values.md))
+
+### Documentation Requirements
+
+* Must include an in-depth `README.md`, including:
+    * Short description of the Chart
+    * Any prerequisites or requirements
+    * Customization: explaining options in `values.yaml` and their defaults
```
axdotl: This item is tricky. I agree, that all configuration options has to be documented.
But I prefer to do the actual documentation in the `values.yaml` itself (see [Ditto config](https://github.com/eclipse/packages/blob/master/charts/ditto/values.yaml) for instance) and only reference to documentation in values.yaml from `README`

From our experience this has the benefit that documentation is close to where configuration is done (with all the resulting advantages).

What do you think?

CONTRIBUTING.md
```diff
@@ -76,3 +76,61 @@ You should, of course, adapt this header to use the specific mechanism for comme
...
+* Must include a short `NOTES.txt`, including:
+    * Any relevant post-installation information for the Chart
+    * Instructions on how to access the application or service provided by the Chart
+
+### Merge Approval and Release Process
+
+All pull requests will be verified by a CI pipeline which runs the linter an tries to install the chart to at least the three most recent minor releases of kubernetes.
+IA maintainer may add "LGTM" (Looks Good To Me) or an equivalent comment to indicate that a PR is acceptable. Any change requires at least one LGTM. No pull requests can be merged until at least one maintainer signs off with an LGTM.
```
axdotl: This has its origin by the [k8s prow bot](https://github.com/kubernetes/test-infra/tree/master/prow#bots-home), which actually merge PRs if a maintainer comment with LGTM.

Since we are not using this bot, I would adjust this point that a maintainer must agree somehow.

CONTRIBUTING.md
sophokles73: That makes a lot of sense to me :-) I'll change it accordingly ...

CONTRIBUTING.md
sophokles73: I'll adapt the wording accordingly

sophokles73:
@axdotl I have added the suggested changes ...

ctron:
I like the content of the PR … I however have two concerns that I would like to discuss:

1) The original focus on this repository was "packages". Now, without any real introduction, we add "Charts" in the main CONTRIBUTING.md file of the repository. I think we should somehow clarify that.
2) We also have a "contributing" section on the proper homepage, which now contains duplicate information: https://www.eclipse.org/packages/contribute/ … I think, rather than splitting up information, we should keep it in a single location. My preferred location would be the homepage. We could keep the administrative part in the git repositories main file, and the more informative on the homepage. And of course, link between each other.

sophokles73:
> My preferred location would be the homepage. We could keep the administrative part in the git repositories main file, and the more informative on the homepage. And of course, link between each other.

Can you elaborate a little on what you consider *administrative* vs *informative* parts? I can then move the *informative* parts to the homepage, if that is what you want ...

ctron:
> > My preferred location would be the homepage. We could keep the administrative part in the git repositories main file, and the more informative on the homepage. And of course, link between each other.
> 
> Can you elaborate a little on what you consider _administrative_ vs _informative_ parts? I can then move the _informative_ parts to the homepage, if that is what you want ...

"Administrative" is more the "Eclipse Foundation Development Process" part. Informative is more "I am user and interested in helping out".

I would assume that the "informative" part comes before the "administrative" part. My assumption being that a user first interacts with the homepage, and the documentation on it. And then, at some point, with the code in the Git repository.

Just to be clear, I am just trying to have a discussion, and brainstorm, for a solution to these two pages that we currently have.

Now the changes in this PR are mostly in the phase when you actually do interact with the code. And the additional section with the gitter channel is helpful in any case. However, reading either of the documents (contributing.md or contributing page), you have no idea that the other one exists. What I would like to prevent is that both side grow organically, and actually divert from each other.

I think that questions like "can I contribute a non-EclipseIoT helm chart as well?" should be answered in the contributing page, rather than the MD file. In doubt, I would put stuff in the contributing page.

One first step could simple be to amend this PR, and add links from each page to the other, to raise the awareness they exists.

Maybe there is another solution to this issue. I am definitely open to suggestions.

sophokles73:
@ctron I have pushed some changes based on your comments. WDYT?

PR #22, opened 2020-01-16
Fixed typo in README.md

patrickse:
Fixed typo in README.md

relm -> helm

Signed-off-by: Patrick Sernetz <patrick@sernetz.com>

### DIFF
--- a/charts/ditto/Chart.yaml
+++ b/charts/ditto/Chart.yaml
@@ -13,7 +13,7 @@ apiVersion: v1
 appVersion: "1.0.0"
 description: A Helm chart for Eclipse Ditto
 name: ditto
-version: 1.0.0
+version: 1.0.1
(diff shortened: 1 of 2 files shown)


ctron:
@monotek The CI check failed because we need a new version. Technically I understand why … but is that really necessary?

thjaeckle:
I would not bother in creating a new micro version - after all it's a bug fix.

axdotl:
As the README is part of the Helm chart, a change on this also requires a new version. Because a change on the readme results in a new hash for the artifact (the packaged chart), this would cause trouble if the version remains the same but the digest has changed.

patrickse:
Ok got it.. I will create a new version tomorrow as soon as I am in the office... sorry just wanted to fix this little typo, got no clue that this would lead to a new version 😉

PR #23, opened 2020-01-17
use kind 0.7.0 in helm ci pipline

monotek:

* use kind 0.7.0 in helm ci pipeline
* removed check for chart changes as its already replaced via workflow config
* added bitnami helm repo to be able to fetch dependency charts from there


.github/ct.yaml
```diff
@@ -15,3 +15,4 @@ check-version-increment: true
 debug: true
 chart-repos:
   - stable=https://kubernetes-charts.storage.googleapis.com/
+  - bitnami=https://charts.bitnami.com
```
ctron: For what reason do we need this additional repository?

.github/workflows/ci.yaml
```diff
@@ -65,10 +57,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v1
       - name: Create kind cluster
-        uses: helm/kind-action@master
+        uses: monotek/kind-action@kind070
```
ctron: Isn't there a chance we get this in the "helm" action any time soon?

monotek: We can wait before merging this so if its merged there i change it back here.

.github/ct.yaml
axdotl: This will be needed by the Hawkbit chart. Actually this change can be removed from this PR.

.github/workflows/ci.yaml
sophokles73: My understanding is that without this change, the CI pipeline doesn't work for Hono and hawkBit, right? If so, then I would rather have this change merged now and then later, once it has made its way into the *official* `helm/kind-action`, change it back. That way we would be able to make some progress with the Hono and hawkBit charts ...

ctron: I was hoping for a quick merge upstream. But yes, let's merge that and make progress with our stuff.

.github/ct.yaml
ctron: Is there a way we can confine this then to the hawkbit chart?

PR #25, opened 2020-01-23
Add the official Eclipse Hono 1.0.3 chart

sophokles73:
This is just a test to see if the CI pipeline succeeds, do not merge.


sophokles73:
@ctron @monotek this PR was just meant as a test to verify if the original/unaltered chart would pass all checks, which it seems to do (apart from some formal modifications to please the linter).
I would therefore rather add the original 1.0.3 chart under its original name (`eclipse-hono`) here as a baseline for all further development. We would then able to distribute the identical chart from both the Eclipse download servers as well as the IoT Packages repository and maybe use this chart as the basis for creating the first version of package zero. The Eclipse Hono project then can decide without any need for hurry, if, when and how they want to move the chart to the IoT Packages project.
If they do, we can make the changes proposed by @ctron as a first PR against the original version, increasing the version number and maybe also renaming the chart to `hono` as well ...

monotek:
Imho the chart should not be published here unter 2 different names. 
This will only confuse users, even when one of the charts is deprecated later.

As mentioned before, when the chart is published here, it will be made available via https://hub.helm.sh/charts?q=eclipse too. 
So people will start using it from there and migrating chart to another name could give them a hard time.
As you can also see the "eclipse-iot" prefix is already in the repo so naming the chart with "eclipse" prefix should be superfluous.

So i guess it's better to keep the "eclipse-hono" chart where it is now and publish it as "hono" chart here, when you're ready to do so.

sophokles73:
> I guess it's better to keep the "eclipse-hono" chart where it is now and publish it as "hono" chart here, when you're ready to do so.

I am afraid that this might introduce quite some delay because in the Hono project we will need to figure out first if and how we want to move the development of the Hono chart to IoT Packages. If we agree to *not* move the chart, then we might need to wait until Hono 1.1.0 before we can add a Hono chart to IoT Packages. I am not sure if we can afford this delay wrt  package zero ...

@ctron @calohmn @dejanb  WDYT?

monotek:
Maybe a temporary copy of your current state as "hono" chart would work too?

sophokles73:
closing in favor of #28

PR #26, opened 2020-01-24
Add chart version change hints

sophokles73:
Based on the discussion on Gitter I think that it seems reasonable to explicitly describe some circumstances that would/should require a change of a chart's version. Some of the more obvious
cases are documented in the contribution guide. These can serve as a basis for further additions and or changes.

This PR is supposed to serve as a straw man for discussing the subject and coming to an agreement on at least a few basic rules.

### DIFF
--- a/homepage/contribute.md
+++ b/homepage/contribute.md
@@ -124,6 +124,21 @@ At least one maintainer needs to explicitly approve the PR before it can be merg
 
 Once the Chart has been merged, a CI job will automatically package and release the Chart in the [Eclipse IoT Packages repository](https://eclipse.org/packages/repository/).
 
+## When to increase a chart's version number?
+
+Based on the assumption that a chart's external interface and behavior is determined by the
+configuration properties it supports, the following (non-exhaustive) table lists some types
+of changes that should be reflected by a corresponding increase of the patch, minor or major
+version of the chart.
+
+| Type of change | Requires increase of |
+| -------------- | -------------------- |
+| Update the `README.md` file | patch version |
+| Chart introduces a property for explicitly configuring an already existing fixed value | patch version |
+| Chart introduces a property for configuring a new feature with a reasonable default value | minor version |
+| Chart renames an existing configuration property | major version |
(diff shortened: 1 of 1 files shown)


ctron:
I am not sure if we should add and enforce this on a global level. I absolutely understand the reason for semantic versioning, and this might be a good guideline. But I also think that we should leave it up to the projects to make that decision on their own.

axdotl:
What do you think of adding a link to Helm [best practices on versioning](https://helm.sh/docs/topics/chart_best_practices/conventions/#version-numbers)? And maybe a short explanation how semver should be/ is done in the package project.

sophokles73:
I had hoped to find some useful information there as well. However, I found the generic SemVer description to be rather confusing than helpful so I thought it might be a good idea to explain the generic concept by means of some specific examples from our problem domain.
When I asked about how semver would need to be applied to Helm chart versions last week on our Gitter channel, nobody seemed to have an answer so I assumed that others would have similar problems of understanding as me ...

sophokles73:
Adding a link to the best practices on version still makes sense, of course, FMPOV.

homepage/contribute.md
```diff
@@ -122,6 +122,20 @@ At least one maintainer needs to explicitly approve the PR before it can be merg
...
+of changes that should be reflected by a corresponding increase of the patch, minor or major
+version of the chart.
+
+| Type of change | Requires increase of |
+| -------------- | -------------------- |
+| Chart introduces a property for explicitly configuring an already existing fixed value | patch version |
+| Chart introduces a property for configuring a new feature with a reasonable default value | minor version |
+| Chart renames an existing configuration property | major version |
+| Chart removes an existing configuration property | major version |
+
```
axdotl: I think it is worth to mention, that changing only the README.md also needs a patch version increase.
As this often is confusing to users.

sophokles73:
@axdotl @monotek WDYT?

monotek:
LGTM

sophokles73:
@ctron @calohmn any thoughts?

PR #31, opened 2020-02-05
Adopt hono 1.1.0 changes

sophokles73:
This PR adds all changes that have been made since Hono 1.0.3 to the chart.


charts/hono/values.yaml
```diff
@@ -23,17 +23,39 @@ amqpMessagingNetworkExample:
...
     svc:
       annotations: {}
       loadBalancerIP:
+    # resources contains the container's 'requests and limits for CPU and memory
+    # as defined by the Kubernetes API.
+    # Refer to https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
+    # for a description of the properties' semantics.
+    resources:
+      requests:
+        cpu: "0.1"
```
ctron: I would suggest to go for more common format of "100m" instead of "0.1". Same for all the other locations.

charts/hono/values.yaml
```diff
@@ -23,17 +23,39 @@ amqpMessagingNetworkExample:
...
+        memory: "64M"
```
ctron: I guess you want to go for `64Mi` rather than `64M`, which is a difference. Which is also the more common used unit IMHO.

sophokles73:
I'll incorporate your proposed changes when I do the update to 1.1.0

PR #36, opened 2020-02-07
[helm ci] Use helm repo for github action

monotek:
* Use helm repo for github action again as https://github.com/helm/kind-action/pull/16 has been merged
* updated k8s versions
* removed k8s 1.11 as storage class install fails there
* fixes node image varname
* added cluster version to step name …

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,18 +49,18 @@ jobs:
           - v1.11.10
           - v1.12.10
           - v1.13.12
-          - v1.14.9
-          - v1.15.6
-          - v1.16.3
-          - v1.17.0
+          - v1.14.10
+          - v1.15.7
+          - v1.16.4
+          - v1.17.2
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


ctron:
:+1: from me … but is there anyone having concerns about removing Kubernetes 1.11?

monotek:
>  from me … but is there anyone having concerns about removing Kubernetes 1.11?

I guess not because the initial plan about the pipeline was to start only with the 3 most recent ones :D

sophokles73:
@monotek @ctron FMPOV we should also remove at least Kubernetes 1.12 and I wouldn't have a problem with removing 1.13 as well ...

monotek:
In gke 1.14 is not even the available default so I'd like to wait with that ;-)

sophokles73:
@thjaeckle are you reviewing and merging this?

thjaeckle:
> @thjaeckle are you reviewing and merging this?

@sophokles73 ok, will do

PR #37, opened 2020-02-07
First shot at the Cloud2Edge package

sophokles73:
This is the bare minimum to install the Hono and Ditto charts as a
single package. A common Ingress and/or NodePort strategy should be
devised as a next step in order to allow access to the endpoints from
the outside.


sophokles73:
@ctron I guess we need to set up the CI pipeline for the *packages* folder as well.

sophokles73:
@thjaeckle I think we need to start talking about a common ingress strategy for Hono and Ditto within this package :-)

packages/cloud2edge/README.md
```diff
@@ -0,0 +1,44 @@
...
+IoT application.
+
+## Installing the Chart
+
+To install the chart with the release name `c2e`, run the following command (tested with Helm v3):
+
+```bash
+helm repo add eclipse-iot https://eclipse.org/packages/charts
+helm repo update
+helm install c2e eclipse-iot/cloud2edge
```
thjaeckle: In order for that to work we would have to host the packages also via the `https://eclipse.org/packages/charts` `index.yaml`.
I would not mix charts of single components with complete packages, that might be confusing for users (and for us maintainers).

I would host them under a separate folder, e.g. `https://eclipse.org/packages/packages` and also publish another Helm Hub alias, e.g. `eclipse-iot-packages`.
That way the "cloud2edge" could be installed via:
```
helm install c2e eclipse-iot-packages/cloud2edge
```

sophokles73:
> I think that this PR should be merged with #38

Sure, I can do that.

> Also I think this PR should include the update to the homepage content.

I do not think that I can follow. Which update are you referring to?

packages/cloud2edge/README.md
sophokles73: > I would not mix charts of single components with complete packages, that might be confusing for users (and for us maintainers).

I have no particular preference for it. In any case, the *packages* folder needs to be included in the CI job. If we go for a separate chart repo, then we also nee to modify the web site build job on Eclipse Jenkins accordingly, right @ctron ?

ctron: > If we go for a separate chart repo, then we also nee to modify the web site build job on Eclipse Jenkins accordingly, right @ctron ?

Yes, if we want to have an additional repository, we also need to adapt the Jenkins job. That shouldn't be a big deal though.

ctron:
> > Also I think this PR should include the update to the homepage content.
> 
> I do not think that I can follow. Which update are you referring to, @ctron ?

The instructions here: https://github.com/eclipse/packages/blob/master/homepage/_packages/package-zero/installation.md

sophokles73:
@ctron I have pushed changes regarding the docs on the web site. However, I wonder if the website is the right place for these docs. For the charts we seem to have chosen the chart's README file to contain all relevant information for installing the chart. Do you consider the packages to require a different approach here?

ctron:
> Do you consider the packages to require a different approach here?

Yes I would. For packages I would see the initial point of contact the website. Giving an overview and then walking people through the architecture and then the installation. The Helm chart is more a convenience.

However, I think it makes sense to point people to the package documentation from the Helm README and the final installation message. Something like "Want to know more? Read here -> https://…"

ctron:
I just checked the homepage. Looks good to me. The only thing I found a bit odd is that sometimes the name is `cloud2edge` and sometimes `cloud-2-edge`. I am fine with either, but would prefer using the same style in all locations.

sophokles73:
> The only thing I found a bit odd is that sometimes the name is cloud2edge and sometimes cloud-2-edge

Can you point out one location where `could-2-edge` is used? A search for that string in my workspace didn't yield any results ...

sophokles73:
> I think it makes sense to point people to the package documentation from the Helm README and the final installation message.

@ctron I believe we agree that this means having the installation instructions in the package chart's README instead of the web page, right?

ctron:
> > I think it makes sense to point people to the package documentation from the Helm README and the final installation message.
> 
> @ctron I believe we agree that this means having the installation instructions in the package chart's README instead of the web page, right?

Sorry, the "package document" is the web page for me. So what I meant is, adding a link from the README of the Helm chart, to the web page, containing the package documentation.

ctron:
> > The only thing I found a bit odd is that sometimes the name is cloud2edge and sometimes cloud-2-edge
> 
> @ctron Can you point out one location where `could-2-edge` is used? A search for that string in my workspace didn't yield any results ...

I think you got it all fixed in 45469ef88e1e4e5dc531f3dcf7c206a6583cf4b0

PR #38, opened 2020-02-07
Remove obsolete package zero

sophokles73:
Now that we have a name for the Hono, Ditto, hawkBit package
(Cloud2Edge), Package Zero is no longer needed.

PR #37 contains a first shot at the Cloud2Edge package.


sophokles73:
merged into #37

PR #39, opened 2020-02-10
Enable CI pipeline for packages

sophokles73:
This should set up the CI pipeline to also test packages

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,6 +15,7 @@ on:
   pull_request:
     paths:
       - 'charts/**'
+      - 'packages/**'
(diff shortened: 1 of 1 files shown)


sophokles73:
@ctron any thoughts?

ctron:
Not sure if that works. I would have expected some check in this PR already. But the only one I see is the eca check.

sophokles73:
> I would have expected some check in this PR already

The CI pipeline that is *live* doesn't cover the *packages* folder and the updated pipeline definition file is not being run (yet). Why don't we simply give it a shot, merge it and see what happens when I update #37 ?

PR #54, opened 2020-02-18
Set up Hono connection and digital twin as part of chart install

sophokles73:
The chart now uses a post-install hook to set up the AMQP 1.0 connection
to Hono and add a digitial twin for the demo-device to Ditto.


homepage/_packages/cloud2edge/tour.md
```diff
@@ -19,38 +19,66 @@ or that you might experience some issues when testing.
...
+{% endclipboard %}
+
+## Retrieving the digital twin's current state
+
+The updated state of the digital twin can then be retrieved using:
+
+{% clipboard %}
+curl -u ditto:ditto http://$DITTO_API_IP:$DITTO_API_PORT_HTTP/api/2/things/org.eclipse.packages.c2e:demo-device
+{% endclipboard %}
+
```
thjaeckle: Additionally to the curl polling we could also add subscribing to the Ditto "SSE" (Server Sent Events) which gets each modification to the twin pushed:
```
{% clipboard %}
curl --http2 -u ditto:ditto -H "Accept:text/event-stream" -N http://$DITTO_API_IP:$DITTO_API_PORT_HTTP/api/2/things
{% endclipboard %}
```

sophokles73: That's even better, great idea :+1:

homepage/_packages/cloud2edge/scripts/setCloud2EdgeEnv.sh
```diff
@@ -0,0 +1,76 @@
...
+getService service-device-registry-ext "http https" REGISTRY
+getService adapter-amqp-vertx "amqp amqps" AMQP_ADAPTER
+getService adapter-coap-vertx "coap coaps" COAP_ADAPTER
+getService adapter-http-vertx "http https" HTTP_ADAPTER
+getService adapter-mqtt-vertx "mqtt secure-mqtt" MQTT_ADAPTER
+getService ditto-nginx "http" DITTO_API
+
+echo "# Run this command to populate environment variables"
+echo "# with the NodePorts of Hono's API endpoints:"
+echo "# eval \"\$(./nodePorts.sh RELEASE_NAME [NAMESPACE])\""
```
ctron: I think the script is named `setCloud2EdgeEnv.sh`.

homepage/_packages/cloud2edge/scripts/setCloud2EdgeEnv.sh
```diff
@@ -0,0 +1,76 @@
...
+getService dispatch-router-ext "amqp amqps" AMQP_NETWORK
+getService service-device-registry-ext "http https" REGISTRY
+getService adapter-amqp-vertx "amqp amqps" AMQP_ADAPTER
+getService adapter-coap-vertx "coap coaps" COAP_ADAPTER
+getService adapter-http-vertx "http https" HTTP_ADAPTER
+getService adapter-mqtt-vertx "mqtt secure-mqtt" MQTT_ADAPTER
+getService ditto-nginx "http" DITTO_API
+
+echo "# Run this command to populate environment variables"
+echo "# with the NodePorts of Hono's API endpoints:"
```
thjaeckle: It's not only about Hono's API endpoints, but API endpoints from the cloud2edge package.

thjaeckle:
> As we already discussed, I would like to see the script running from inside a container on the cluster, so that people don't need to download the script. But that can be done in a future iteration.

As I understood the PR this is exactly what is done.
The script does not need to be executed by people themselves, but this is automatically done in `post-install-job.yaml` as Helm post install hook.

sophokles73:
> The script does not need to be executed by people themselves, but this is automatically done in post-install-job.yaml as Helm post install hook.

I think @ctron is talking about the shell script for populating the environment vars with the endoint IPs/ports ...

PR #60, opened 2020-03-03
eclipse- prefix in service names

soheilade:
currently all service names start with eclipse-, I updated readme accordingly.
Signed-off-by: SOHEILA DEHGHANZADEH <s.dehghanzadeh@denso-auto.de>


ctron:
Thank you @soheilade for the PR. I think it looks good, but there are two formalities missing:
* You need it increment the char version, a +1 on the micro version part should be fine
* You need to sign the ECA, and add the `Signed-off-by:` git footer with the e-mail address you signed the ECA for -> https://www.eclipse.org/legal/ECA.php … you will also need to update and force push the current commit.

Let me know if you need any help with this.

PR #62, opened 2020-03-04
added kubeval to github actions pipeline

monotek:
kubeval is a tool for validating a Kubernetes YAML or JSON configuration file. It does so using schemas generated from the Kubernetes OpenAPI specification, and therefore can validate schemas for multiple versions of Kubernetes.

changes:
* added kubeval to github actions pipeline
* removed rm from lint-scripts.sh (is removed anyway)
* added chart-dirs option to ct.yaml to use packages dir too

info: 
* https://github.com/instrumenta/kubeval
* https://kubeval.instrumenta.dev/

### DIFF
--- a/.github/kubeval.sh
+++ b/.github/kubeval.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+#
+# use kubeval to validate helm generated kubernetes manifest
+#
+
+# Copyright (c) 2019 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0
+#
+# SPDX-License-Identifier: EPL-2.0
+
+set -o errexit
+set -o pipefail
+
+CHART_DIRS="$(git diff --find-renames --name-only "$(git rev-parse --abbrev-ref HEAD)" remotes/origin/master -- charts packages | grep '[cC]hart.yaml' | sed -e 's#/[Cc]hart.yaml##g')"
+HELM_VERSION="v3.1.1"
+KUBEVAL_VERSION="0.14.0"
+SCHEMA_LOCATION="https://raw.githubusercontent.com/instrumenta/kubernetes-json-schema/master/"
+
+# install helm
+curl --silent --show-error --fail --location --output get_helm.sh https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get
+chmod 700 get_helm.sh
+./get_helm.sh --version "${HELM_VERSION}"
+
+# install kubeval
+curl --silent --show-error --fail --location --output /tmp/kubeval.tar.gz https://github.com/instrumenta/kubeval/releases/download/"${KUBEVAL_VERSION}"/kubeval-linux-amd64.tar.gz
+sudo tar -C /usr/local/bin -xf /tmp/kubeval.tar.gz kubeval
+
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -40,10 +40,25 @@ jobs:
...
+  kubeval-chart:
+    runs-on: ubuntu-latest
+    needs: check-for-chart-changes
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Set user permissions
+        run: chmod -R 777 /home/runner/work
+      - name: Run kubeval
+        uses: docker://kiwigrid/gcloud-kubectl-helm
```
ctron: Could you explain what this is doing?

.github/kubeval.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
ctron: It would be good to have an understanding of the copyright of this file. I am assuming you are the original author of it.

monotek:
> Could you explain what the goal of this PR is.

added it to description

.github/workflows/ci.yaml
monotek: > Could you explain what this is doing?

checks out repo
set file permissions whats a workaround for wrong /home dir in github actions
uses kiwigrids gcloud-kubectl-helm docker image which contains kubeval binary to validate the helm template

.github/kubeval.sh
monotek: added license header

.github/workflows/ci.yaml
ctron: I would like to avoid adding additional container images. Especially as this isn't versioned.

Why not pull in the binary from https://github.com/instrumenta/kubeval/releases with `curl` instead?

monotek: because its not only the kubeval binary you need.
helm binary is also needed and has to be initialised.
added the docker tag now.

ctron: So to me this sounds more like something we should then add to the packages project. As it wouldn't be possible for IoT Packages committers to make any changes to this docker image in the future.

monotek: i've changed the pr to not use docker.

ctron:
Two things came into my mind, which I didn't think of in the first place:

* Which version of Kubernetes does this validate for?
* Will it still be possible to use CRDs, which are not known to kubeval?

monotek:
> * Which version of Kubernetes does this validate for?

I've added an k8s version matrix now, to validate against same major versions as the chart is installed in kind. Minor versions defer at the moment as https://github.com/instrumenta/kubernetes-json-schema does not contain all versions we use for kind. I'll try to update the repo via pr later.

> * Will it still be possible to use CRDs, which are not known to kubeval?

Yes. I've dded the "--ignore-missing-schemas" flag.

ctron:
> > * Which version of Kubernetes does this validate for?
> 
> I've added an k8s version matrix now, to validate against same major versions as the chart is installed in kind. Minor versions defer at the moment as https://github.com/instrumenta/kubernetes-json-schema does not contain all versions we use for kind. I'll try to update the repo via pr later.

By major you mean minor? And by minor you mean micro/service/patch? If that is the case, then I think we are good.

> 
> > * Will it still be possible to use CRDs, which are not known to kubeval?
> 
> Yes. I've dded the "--ignore-missing-schemas" flag.

Thx!

monotek:
yep, i've mixed that up, sorry...

ctron:
> yep, i've mixed that up, sorry...

no worries, I just wanted to ensure I did not miss the release of K8S 2.0 :grin:

PR #63, opened 2020-03-05
[#61] Increase Hono data grid readinessProbe timeout

### LINKED ISSUE #61: Hono: Data-grid pod doesn't get ready
I've deployed the Hono chart in a [kind](https://kind.sigs.k8s.io/) cluster using the `dataGridExample.enabled=true` option to deploy an Infinispan data grid for the Hono device connection service.

The `eclipse-hono-data-grid-0` pod didn't get ready here, also causing the protocol adapter pods to not get ready.
Error in the `eclipse-hono-data-grid-0` pod:
`Readiness probe errored: rpc error: code = DeadlineExceeded desc = failed to exec in container: timeout 3s exceeded: context deadline exceeded`.

When running the readiness probe command ([`/usr/local/bin/is_healthy.sh`](https://github.com/jboss-dockerfiles/infinispan/blob/9.4.x/server/is_healthy.sh)) manually in the pod, the command took 22s - so way more than the 3s timeout defined in the chart.

I think the timeout should be increased here. Better yet would be to update to Infinispan >= 10, where there is a proper [readinessProbe REST endpoint](https://github.com/infinispan/infinispan-images#liveness-and-readiness-probes).

calohmn:
This fixes #61.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.1.5
+version: 1.1.6
 # Version of Hono being deployed by the chart
 appVersion: 1.1.1
 keywords:
--- a/charts/hono/templates/example-data-grid/statefulset.yaml
+++ b/charts/hono/templates/example-data-grid/statefulset.yaml
@@ -63,11 +63,11 @@ spec:
           exec:
             command:
             - /usr/local/bin/is_healthy.sh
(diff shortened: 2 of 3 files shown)


calohmn:
@sophokles73 I don't think the `-XX:MinRAMPercentage=80 -XX:MaxRAMPercentage=80` options matter here because in the script `-Xmx16m` is set, and that gets precedence (see [here](https://stackoverflow.com/a/54297753)).

> Anyway, I guess we could fall back to simply using an HTTP based probe on Infinispan's REST API instead of the Java based is_healthy.sh ...

Right, I had been concentrating on the fact that a dedicated health HTTP endpoint is only available from Infinispan 10 onwards, but if you have another command ready, we can use that. I can close this PR then.

PR #66, opened 2020-03-05
Configmap for Nginx should not be deploy when nginx is disabled

BobClaerhout:

### DIFF
--- a/charts/ditto/templates/nginx-configmap.yaml
+++ b/charts/ditto/templates/nginx-configmap.yaml
@@ -9,6 +9,7 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 ---
+{{- if .Values.nginx.enabled -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -162,3 +163,5 @@ data:
         {{- end }}
       }
     }
+
+{{- end }}
--- a/charts/ditto/Chart.yaml
+++ b/charts/ditto/Chart.yaml
@@ -17,7 +17,7 @@ description: |
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids,
   EV charging stations etc).
 name: ditto
-version: 1.1.3
+version: 1.1.4
 keywords:
   - iot-chart
 home: https://www.eclipse.org/ditto


charts/ditto/templates/nginx-configmap.yaml
```diff
@@ -162,3 +163,5 @@ data:
         {{- end }}
       }
     }
+
+{{- end }}
```
sophokles73: missing EOL

sophokles73:
@BobClaerhout you also need to bump the (patch) version number ...

thjaeckle:
@BobClaerhout could you please merge the latest master into your branch?
I updated the GitHub "checkout" action to v2 - this version fixes that failed CI could not be retried.

BobClaerhout:
> @BobClaerhout could you please merge the latest master into your branch?
> I updated the GitHub "checkout" action to v2 - this version fixes that failed CI could not be retried.

Done. Tests still failing though

thjaeckle:
I don't see any error message why this now fails.
Anyone else can explain that?

ctron:
Me neither @monotek do you have an idea?

charts/ditto/templates/nginx-configmap.yaml
```diff
@@ -9,6 +9,7 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 ---
+{{- if .Values.nginx.enabled -}}
```
sophokles73: I am not sure if the `-` in `-}}` is correct. FMPOV it will lead to `apiVersion: v1` being pulled up directly behind the `---` delimiter ...
@BobClaerhout can you check by running `helm template` locally?

BobClaerhout: I do not get any errors when running `helm template` locally. 
Also, this is a copy from https://github.com/eclipse/packages/blob/a21fb8306e1bc2116efeda6c02cb023ba2960ec5/charts/ditto/templates/nginx-config.yaml#L12 so this should be ok.

ctron:
I am having the same problem on #73 … @monotek I think there is something broken in the github action, could that be? As we are using "master", this might have changed to something which is no longer compatible.

ctron:
As mentioned in https://github.com/helm/chart-testing-action/issues/20#issuecomment-598515402 … this a problem with `actions/checkout@v2`.

@thjaeckle did introduce v2 in commit a6f7c8824147cb06092e3bb9d37296bd899bb109

I did revert that now, and indeed, linting works again.

So you need to rebase your PR on master, and push again. That should fix the issue.

thjaeckle:
Damn. thx @ctron
Did update to v2 as triggering rebuilds of failed GH Actions checks does not work with v1.
Had the same problem in our Ditto actions.
But there we don't use chart testing.

ctron:
> Did update to v2 as triggering rebuilds of failed GH Actions checks does not work with v1.

Which means we can only have **either** triggering re-builds **or** linting :roll_eyes:

Hopefully this will get fixed in helm linting soon.

PR #67, opened 2020-03-05
Support connecting external protocol adapters

sophokles73:
The example AMQP Messaging Network and Device Registry can now be
configured to allow protocol adapters that run outside of the cluster to
connect to their secure endpoints.
This is helpful e.g. for connecting edge nodes which run a Hono protocol
adapter to connect to a Hono instance running in a back end.

### DIFF
--- a/charts/hono/config/router/qdrouterd.json
+++ b/charts/hono/config/router/qdrouterd.json
@@ -41,7 +41,6 @@
     "requireSsl": true,
     "host": "0.0.0.0",
     "port": 5671,
-    "maxSessionFrames": 200,
     "linkCapacity": 100,
     "authenticatePeer": true,
     "saslMechanisms": "PLAIN",
@@ -51,7 +50,6 @@
   ["listener", {
     "host": "0.0.0.0",
     "port": 5672,
-    "maxSessionFrames": 200,
     "linkCapacity": 100,
     "authenticatePeer": true,
     "saslMechanisms": "PLAIN",
@@ -63,7 +61,6 @@
     "requireSsl": true,
     "host": "0.0.0.0",
     "port": 5673,
-    "maxSessionFrames": 500,
     "linkCapacity": 100,
     "authenticatePeer": true,
     "saslMechanisms": "EXTERNAL"
@@ -135,6 +132,7 @@
       "groups": {
         "$default": {
           "remoteHosts": "*",
+          "maxSessionWindow": 3276800,
           "maxSessions": 10
         }
       }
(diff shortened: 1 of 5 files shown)


charts/hono/config/router/qdrouterd.json
```diff
@@ -148,11 +146,12 @@
           "users": "Eclipse IoT;Hono",
           "remoteHosts": "*",
           "maxSessions": 4,
+          "maxSessionWindow": 8192000,
           "maxMessageSize": 131072,
           "allowUserIdProxy": true,
           "allowAnonymousSender": true,
-          "sources": "control/*, command/*",
-          "targets": "telemetry/*, event/*, control/*, command_response/*"
```
calohmn: The `control/*` addresses are still needed as of now (Hono 1.1).
(In Hono 1.2, that pattern will be renamed as part of the changes for https://github.com/eclipse/hono/issues/1272.)

charts/hono/values.yaml
```diff
@@ -182,6 +182,24 @@ useLoadBalancer: true
 # Configuration properties for protocol adapters.
 adapters:
 
+  # externalAdaptersEnabled indicates whether protocol adapters that
+  # run outside of the kubernetes cluster should be allowed to connect to
+  # the Dispatch Router and the example device registry's service endpoints.
+  # The default setting is 'false' which prevents access from outside of the
+  # cluster.
+  # Setting this property to 'true' allows external adapters to connect to
+  # the Dispatch Router's 'interal' endpoint and the Device Registry's
```
calohmn: typo "interal"

charts/hono/config/router/qdrouterd.json
sophokles73: This is still used for forwarding commands that need to be mapped to a gateway ID, right?

charts/hono/values.yaml
```diff
@@ -182,6 +182,24 @@ useLoadBalancer: true
...
+  # service endpoints via AMQPS, i.e. AMQP over TLS.
+  # The Dispatch Router requires adapters to authenticate using SASL EXTERNAL,
+  # i.e. an adapter needs to provide a client certificate that has been signed
+  # by one of the CA certs contained in the router's trust store.
+  # When opening the AMQP connection to the router, the adapter needs to indicate
+  # the 'hono-internal' virtual host name in its AMQP 1.0 'open' frame.
```
calohmn: I think a hint would be helpful that the connection has to be opened to a different port than the one to be used by northbound applications when connecting to the Dispatch Router.

charts/hono/config/router/qdrouterd.json
calohmn: yes

sophokles73:
@calohmn I have incorporated your feedback. Would you mind taking another look?

PR #73, opened 2020-03-09
Add routes and profile for OpenShift

ctron:


ctron:
@calohmn @laverman could you take a look?

PR #75, opened 2020-03-18
Add support for mounting additional secrets in broker and router

sophokles73:
The example registry's Dispatch Router and Artemis containers can now be
configured to mount additional secret volumes. This way, the router and
broker can be configured with custom key material.
The example regstry's Dispatch Router's "internal" listener can now be
configured with the uidFormat and to use for extracting user IDs from
client certificate subject DNs. This is needed in order to properly
configure the protocol adapter uids when using custom certificates from
your own PKI.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.2.1
+version: 1.3.0
 # Version of Hono being deployed by the chart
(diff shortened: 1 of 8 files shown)


sophokles73:

sophokles73:
@ctron do you have any clue why the jenkins pr-merge check would fail?

ctron:
> @ctron do you have any clue why the jenkins pr-merge check would fail?

Yes, we can ignore that. It was a problem in the configuration of the Jenkins GitHub plugin. It seems there is a newer version, which was auto-configured to process PRs. However we are not set up to process PRs with Jenkins and so I disabled this part. However Jenkins already marked PRs failing.

When you push some new commit (change timestamp and force push) this check will no longer be triggered, and thus no longer fail.

Issue #74, opened 2020-03-12
Upload does not work, storage not configured


ctron:
Performing the default installation of Hawkbit with the chart will result in an installation, which cannot handle uploads.

It seems that the artifact gets uploaded to some local directory `./artifactrepo`. However that doesn't exists and so the upload fails.

In order to fix this, I had to create a PVC and then add the following values:

~~~yaml
extraEnv:
  "org.eclipse.hawkbit.repository.file.path": /var/lib/hawkbit-storage
extraVolumes:
  - name: storage
    persistentVolumeClaim:
      claimName: hawkbit-storage
extraVolumeMounts:
  - name: storage
    mountPath: /var/lib/hawkbit-storage
~~~

I think there should be an out of the box PVC. Otherwise hawkbit can't really do much.

laverman:
As an alternative, we could utilize the [MongoDB extension of hawkBit](https://github.com/eclipse/hawkbit-extensions/tree/master/hawkbit-extension-artifact-repository-mongo) to store the uploaded artifacts there. I think this would be the cleaner solution. However, this would further increase the resource consumption of the stack (or is there already a MongoDB in place?). 
What do you think?

thjaeckle:
A MongoDB is already in place and started via the Ditto Helm chart.
If we want to use a shared one, we probably would need to add it as dependency to the "packages", disable the MongoDB dependency in the Ditto chart and configure Ditto to use the one from the package.
But that should work.

ctron:
> As an alternative, we could utilize the [MongoDB extension of hawkBit](https://github.com/eclipse/hawkbit-extensions/tree/master/hawkbit-extension-artifact-repository-mongo) to store the uploaded artifacts there. I think this would be the cleaner solution. However, this would further increase the resource consumption of the stack (or is there already a MongoDB in place?).
> What do you think?

I think adding both would be a good idea. Fixing the filesystem one for a simple Hawkbit only deployment. And MongoDB something more advanced. In the c2e package we could then activate MongoDB out of the box.

laverman:
Sounds reasonable. However, would that mean, that Ditto would also support both use-cases: For a Ditto-Only deployment it would spawn its own MongoDB, while in the c2e package, the central instance would be used?

PR #76, opened 2020-03-19
[#74]: Allow using file storage, enable by default

ctron:


charts/hawkbit/templates/pvc.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.fileStorage.enabled -}}
```
sophokles73: how about adding a license header?

ctron: None of the files do have a license header in the hawkbit package. I agree it would be an idea to add them. But I would then do it in another PR, outside of the scope of this one.

PR #77, opened 2020-03-20
Add c2e icon to helm chart

ctron:


packages/cloud2edge/Chart.yaml
```diff
@@ -11,7 +11,7 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.0.5
+version: 0.0.6
```
sophokles73: since there is no single app, we have so far kept the version and appVersion in sync. Otherwise I wonder, which application (Hono, Ditto, hawkBit) the *appVersion* refers to ...

ctron: hm, good point. changed.

PR #79, opened 2020-03-23
Upon restart, file base device registry configuration is overwritten

BobClaerhout:
Somehow, this slipped back in: https://github.com/eclipse/hono/issues/1007

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.3.0
+version: 1.3.1
 # Version of Hono being deployed by the chart
 appVersion: 1.1.1
 keywords:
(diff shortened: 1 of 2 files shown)


ctron:
@BobClaerhout could you rebase on master, that should fix the CI issues

BobClaerhout:
> @BobClaerhout could you rebase on master, that should fix the CI issues

Oops, sorry. Should have known it from the first time.
Done

BobClaerhout:
No changes anymore from master to the current branch.

sophokles73:
@BobClaerhout this did not *slip back in* but has been fixed again so that the copy commands can actually be run by the busybox image. You can try for yourself:
```
 $ docker run -it --rm busybox /bin/sh -c 'cp -n /etc/hostname /etc/hostname.bak'
cp: invalid option -- n
BusyBox v1.31.1 (2020-03-09 21:47:38 UTC) multi-call binary.

Usage: cp [OPTIONS] SOURCE... DEST

Copy SOURCE(s) to DEST

	-a	Same as -dpR
	-R,-r	Recurse
	-d,-P	Preserve symlinks (default if -R)
	-L	Follow all symlinks
	-H	Follow symlinks on command line
	-p	Preserve file attributes if possible
	-f	Overwrite
	-i	Prompt before overwrite
	-l,-s	Create (sym)links
	-T	Treat DEST as a normal file
	-u	Copy only newer files
```
The included version of *cp* simply doesn't support the `-n` option ...

BobClaerhout:
Ok, but the problem still exists if we keep the `cp -u`. The data gets overwritten upon each run of the busybox.

BobClaerhout:
Updated with a check in bash

ctron:
Ok, the CI is happy again.

PR #84, opened 2020-03-31
Update README.md

rainbowdn:
typo fix


ctron:
Thanks for the PR. The chart version needs to be incremented, even though it is a simple typo fix.

PR #87, opened 2020-04-01
[#78] Configure Hawkbit to accept proxy forward headers by default

### LINKED ISSUE #78: Hawkbit hands out internal URLs
Deploying Hawkbit using the Helm chart, putting it behind a proxy, makes hawkbit return wrong Urls.

Hawkbit is reachable as:

    https://hawkbit.foo.bar/

But returns the following URL (note the missing `s` in `http`):

    http://hawkbit.foo.bar/DEFAULT/controller/v1/dev01/configData

ctron:
Closes #78


charts/hawkbit/values.yaml
```diff
@@ -123,6 +123,8 @@ spring:
 
 config:
   application:
+    server:
+      useForwardHeaders: true
```
vigier, line 127: Not sure if it is a good idea to turn this option on by default, as it implies that some tricky security consideration have been made, i.e. that you trust the proxy where the requests are coming from _and_ that you really running behind a proxy. If there is no validation for the forwarded host in place you might become vulnerable for a [host-header attack](https://portswigger.net/web-security/host-header) - and to be hones - I have not yet verified that hawkBit + embedded server is doing a verification.
I have read that the useForwareHeaders is [switched on by default](https://docs.spring.io/spring-boot/docs/2.0.x/reference/htmlsingle/#howto-use-tomcat-behind-a-proxy-server) when deployed in cloud-foundry or heroku... maybe there is a way to do the same with K8s?

laverman (approved):

PR #95, opened 2020-04-09
update readme of c2e with link to homepage

rainbowdn:


ctron:
Thanks for the PR.

My concern is, that we are duplicating the information, which should already be here: https://github.com/eclipse/packages/blob/master/homepage/_packages/cloud2edge/installation.md

I would prefer if the `README.md` would simply point to the rendered version of this file: https://www.eclipse.org/packages/packages/cloud2edge/installation/

So that we have all the information in one place. And so that we can also provide the different variants, which the homepage provides.

sophokles73:
@rainbowdn can you please rebase and resolve the conflict so that we can merge this PR?

packages/cloud2edge/Chart.yaml
```diff
@@ -12,7 +12,7 @@
 #
 apiVersion: v1
 version: 0.0.7
-appVersion: 0.0.7
+appVersion: 0.0.8
```
sophokles73, line 15: you need to change both, the *version* as well as the *appVersion*

sophokles73:
@rainbowdn please *rebase* on *HEAD* of the master branch

PR #101, opened 2020-04-22
Add mapper endpoints

BobClaerhout:


BobClaerhout:
Should I do a version bump? If so, to what version?

ctron:
Are the mapper endpoints available in Hono 1.2 already?

BobClaerhout:
> Are the mapper endpoints available in Hono 1.2 already?

No, but didn't know where to do this else? Maybe this should be here until 1.3 is released? How does this work to keep them in sync?

ctron:
> > Are the mapper endpoints available in Hono 1.2 already?
> 
> No, but didn't know where to do this else? Maybe this should be here until 1.3 is released? How does this work to keep them in sync?

I think it is ok to keep this PR open, maybe we need a specific label for that (like "wait for upstream").

Maybe it also makes sense to create a tracking issue in Hono, referencing this PR.

sophokles73:
@BobClaerhout just realized that this has been lingering around for quite some time. Sorry for that.
Would you mind rebasing this PR and bumping the Chart's version so that the CI build picks it up?

BobClaerhout:
> @BobClaerhout just realized that this has been lingering around for quite some time. Sorry for that.
> Would you mind rebasing this PR and bumping the Chart's version so that the CI build picks it up?

updated it

charts/hono/templates/hono-adapter-mqtt/hono-adapter-mqtt-vertx-secret.yaml
```diff
@@ -33,6 +33,10 @@ stringData:
         certPath: /etc/hono/cert.pem
         tenantIdleTimeout: {{ .Values.adapters.defaultTenantIdleTimeout | quote }}
         {{- end }}
+        {{- if .Values.adapters.mapperEndpoints }}
```
sophokles73, line 36: this will always evaluate to `null` FMPOV as the corresponding property is defined at path
`adapters.mqtt.mapperEndpoints` in the values.yaml file. I do not think that we need to define the mapper endpoints per adapter type because we already talked about having this as a general feature available to all adapters, right? So I would rather update the values.yaml file accordingly and move the property right beneath the `adapters` property.
WDYT?

BobClaerhout: good catch. updated the `values.yaml` file

sophokles73:
@BobClaerhout can you bump the version to 1.4.5 please?

sophokles73:
@BobClaerhout this didn't seem to have worked, can you rebase/resolve the conflict?

ctron (approved):

PR #104, opened 2020-04-30
configured c2e package to use Ditto 1.1.0

thjaeckle:
by using Ditto chart ~1.2.0

### DIFF
--- a/packages/cloud2edge/Chart.yaml
+++ b/packages/cloud2edge/Chart.yaml
@@ -11,8 +11,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.0.9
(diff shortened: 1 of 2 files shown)


packages/cloud2edge/Chart.yaml
```diff
@@ -11,8 +11,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.0.8
-appVersion: 0.0.8
+version: 0.0.9
+appVersion: 0.0.9
```
sophokles73, line 15: this will need to be updated to 0.0.10 with the latest merge ...

thjaeckle: done

PR #111, opened 2020-06-25
[eclipse/hono#1679] Add support to deploy MongoDB based device registry

kaniyan:
With this PR, the Hono charts extend support to deploy _MongoDB_ based device registry and also a _MongoDB_ instance (if required). The users can now easily choose if the _file_ or _MongoDB_ based device registry is to be deployed.


charts/hono/values.yaml
```diff
@@ -823,107 +823,247 @@ deviceRegistryExample:
...
-      # using an example key and certificate.
-      rest:
-      #  insecurePortEnabled: true
-      #  insecurePortBindAddress: "0.0.0.0"
+  # mongoDBBasedDeviceRegistry contains configuration properties for the
+  # example MongoDB based Device Registry.
+  mongoDBBasedDeviceRegistry:
+    # enabled indicates whether the example MongoDB based Device Registry
+    # should be deployed and used.
+    enabled: false
```
sophokles73, line 939: this basically means that we can deploy both registries, the file based and the mongo DB based together, right?
how about using a single property *type* with possible values `file` and `mongo` instead?

charts/hono/values.yaml
```diff
@@ -823,107 +823,247 @@ deviceRegistryExample:
...
+        # rest contains configuration properties for the server's
+        # exposed HTTP endpoints.
+        # If not set, the registry by default exposes the insecure and secure ports
+        # using an example key and certificate.
+        http:
+        #  insecurePortEnabled: true
+        #  insecurePortBindAddress: "0.0.0.0"
+      # mongodb contains the configuration properties to connect to the MongoDB database instance.
+      # If you would like to use an already existing MongoDB database instance, then configure
+      # the below section accordingly.
```
sophokles73, line 991: it looks like all of this is the same as for the file based registry, I'd rather have a single set of properties to configure the endpoints in order to make it less error prone and keep the properties and comments consistent ...

kaniyan: Not all are same as of the file based registry. For example it is `http` in MongoDB registry and in case of file-based it is `rest`. And MongoDB registry  has some extra properties (with prefix `mongodb`)  to configure the database connection. In future we will have more (eg. JDBC based) device registry implementations and putting all properties under one section IMHO would be difficult and not easy to configure. Also I thought it may be easier for users who wants to use either MongoDB or file-based registry to focus on individual section, makes changes there and need not worry about the other registries. WDYT?

sophokles73: They will all expose endpoints via HTTP and AMQP, won't they? And they will all need to authenticate clients on the AMQP endpoint using the Authentication service. So there is little need to repeat these config properties for all of the registries, is there?

kaniyan: I see your point to reduce the repetition and it has its advantages too. Just to ensure that my understanding is correct, the idea is to have one common section for configuring the properties that are common. The rest of the properties such as `data`, image details will be in separate sections (MongoDb and file) as below.

```
deviceRegistryExample:
  enabled: true
  type: mongodb / file
  javaOptions: -XX:MinRAMPercentage=80 -XX:MaxRAMPercentage=80
  hono:
    auth:
    registry:
      amqp:
      rest:
  resources:
    requests:
      cpu: "100m"
      memory: "300Mi"
    limits:
      cpu: "1"
      memory: "300Mi"
  svc:
    annotations: {}
    loadBalancerIP:
  healthCheck:

  fileBasedDeviceRegistry:
    imageName: index.docker.io/eclipse/hono-service-device-registry-file
    data:
      volumeName:
      tenantsFile: example-tenants.json
      devicesFile: example-device-identities.json
      credentialsFile: example-credentials.json
      resetOnStartup: true
    storageClass:

  mongoDBBasedDeviceRegistry:
    imageName: index.docker.io/eclipse/hono-service-device-registry-mongodb
    hono:
      mongodb:
```

charts/hono/values.yaml
ctron: and `jdbc` in the future

kaniyan: > and `jdbc` in the future

This PR takes into account that as many types along with that type specific configurations be added in the future. When the _jdbc_ based device registry is there, it can be easily added as a new type along with _jdbc_ specific configuration.

charts/hono/values.yaml
```diff
@@ -925,6 +899,105 @@ deviceRegistryExample:
...
+      resetOnStartup: true
+    # storageClass indicates the name of the kubernetes StorageClass that
+    # should be used for the example registry's persistent volume claim.
+    # If not set, the cluster's default storage class is used.
+    storageClass:
+
+  # mongoDBBasedDeviceRegistry contains configuration properties specific to the
+  # MongoDB based device registry.
+  mongoDBBasedDeviceRegistry:
+    # imageName contains the name (including registry and tag)
```
sophokles73, line 940: it doesn't contain the tag, does it?

charts/hono/values.yaml
```diff
@@ -925,6 +899,105 @@ deviceRegistryExample:
     #  keyPath: "/etc/config/key.pem"
     #  certPath: "/etc/config/cert.pem"
 
+  # fileBasedDeviceRegistry contains configuration properties specific to the
+  # file based device registry.
+  fileBasedDeviceRegistry:
+    # imageName contains the name (including registry and tag)
```
sophokles73, line 905: it doesn't contain the tag, does it?

sophokles73 (commented):
all of the data used to initialize the DB with tenant, devices, credentials is the same as for the file based registry. IMHO we should try to have the definition of the data in a single place only and then use it for both the file based as well as the Mongo DB based registries. Otherwise we will have a hard time keeping it all consistent ...

kaniyan:
> all of the data used to initialize the DB with tenant, devices, credentials is the same as for the file based registry. IMHO we should try to have the definition of the data in a single place only and then use it for both the file based as well as the Mongo DB based registries. Otherwise we will have a hard time keeping it all consistent ...

Absolutely, I agree with it. I would suggest to address that in a separate PR, so that users can asap start using Hono charts to deploy also MongoDB device registry. I will create an issue to track it.

charts/hono/values.yaml
kaniyan: I have change it now and pushed a commit to address it and also the other comments. Would you mind taking a look again?

ctron:
> > all of the data used to initialize the DB with tenant, devices, credentials is the same as for the file based registry. IMHO we should try to have the definition of the data in a single place only and then use it for both the file based as well as the Mongo DB based registries. Otherwise we will have a hard time keeping it all consistent ...
> 
> Absolutely, I agree with it. I would suggest to address that in a separate PR, so that users can asap start using Hono charts to deploy also MongoDB device registry. I will create an issue to track it.

I would prefer this to be done in the initial PR. Otherwise we are only piling up technical debt, and later forget about it.

kaniyan:
With the last commit, I have consolidated the example device registry data irrespective of the type of device registries being deployed. This is done by executing a script with curl commands using a _post-install_ hook. @ctron @sophokles73 Would you mind taking a look?

sophokles73:
@kaniyan can you rebase and resolve conflicts? I will then review again ...

kaniyan:
> @kaniyan can you rebase and resolve conflicts? I will then review again ...

@sophokles73 Rebased.

charts/hono/requirements.yaml
```diff
@@ -19,3 +19,7 @@ dependencies:
     version: ~3.8.15
     repository: "https://kubernetes-charts.storage.googleapis.com/"
     condition: grafana.enabled
+  - name: mongodb
+    repository: "https://charts.bitnami.com/bitnami"
+    version: ~7.14.7
```
sophokles73, line 24: which version of MongoDB does this use?

charts/hono/templates/hono-service-device-registry-base/hono-service-device-registry-amqps-route.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if and .Values.deviceRegistryExample.enabled ( eq .Values.platform "openshift" ) }}
 #
-# Copyright (c) 2019 Contributors to the Eclipse Foundation
+# Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
```
sophokles73, line 3: if you do not change anything in the file then you do not need to update the copyright header FMPOV

charts/hono/templates/hono-service-device-registry-base/headless-svc.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.deviceRegistryExample.enabled }}
+{{- if and .Values.deviceRegistryExample.enabled ( eq .Values.deviceRegistryExample.type "file" ) }}
```
sophokles73, line 1: Since it is only used for the file based registry, this file should probably be moved to `hono-service-device-registry-file`, shouldn't it?

sophokles73 (commented):
looks very good :+1:

charts/hono/templates/hono-service-device-registry-base/hono-service-device-registry-amqps-route.yaml
kaniyan: I added some changes and after refactoring I have removed it but didn't update he copyright header. I will revert it.

sophokles73: Note that this affects multiple files ...

charts/hono/requirements.yaml
kaniyan: It uses MongoDB version 4.2.8. It was the latest version of charts at the time of starting with helm charts for MongoDB device registry. Is the motive of the question to align with the version 4.0.20 that we use for the tests? I checked and there are no helm charts for version 4.0.20. Chart version 7.8.0 uses MongoDB version 4.0.14 and the next version 7.8.1 uses MongoDB version 4.2.2.

image

Using a stateful set to deploy as a stand-alone instance has been introduced in version 7.8.0 but it has a bug and it has been fixed in version 7.8.7 but it uses MongoDB version 4.2.3.

charts/hono/templates/hono-service-device-registry-base/hono-service-device-registry-amqps-route.yaml
kaniyan: Thanks for pointing out and I have updated those files and pushed a commit.

charts/hono/requirements.yaml
sophokles73: Yes, I thought that it might be good to use the same version in the chart that we also use in the unit and integration tests, wouldn't you agree? However, this doesn't necessarily mean that we must use 4.0.20 in the chart. We might as well upgrade to a more recent version (4.2.x) in the unit/integration tests. WDYT?

sophokles73:
@kaniyan can you also rebase?

charts/hono/requirements.yaml
kaniyan: I agree that it is good to use the same version. Yes, it is also better that we upgrade to a more recent version in our unit/integration tests.

kaniyan:
> @kaniyan can you also rebase?

@sophokles73 done.

sophokles73 (approved):
LGTM

sophokles73:
@kaniyan one more thing: we need a profile for using the MongoDB in the CI tests run for the chart. For that, you need to create a separate yaml file in the `ci` folder analogously to the existing profile files that are there already ...

charts/hono/ci/mongodb-based-device-registry-values.yaml
```diff
@@ -0,0 +1,31 @@
...
+mongodb:
+  createInstance: true
+
+useLoadBalancer: false
+
+prometheus:
+  createInstance: false
+
+grafana:
+  enabled: false
```
sophokles73, line 31: missing EOL

sophokles73 (changes requested):

charts/hono/ci/mongodb-based-device-registry-values.yaml
```diff
@@ -0,0 +1,31 @@
...
+# profile for installing Hono
+# - with MongoDB based device registry
+# - with MongoDB instance
+# - without monitoring infrastructure
+
+deviceRegistryExample:
+  type: mongodb
+
+mongodb:
+  createInstance: true
```
sophokles73, line 23: IMHO we need to add
```yaml
mongodb:
  persistence:
    enabled: false
```
here in order for the install test to succeed because the chart-testing action does not support PVCs ...

sophokles73 (changes requested):

charts/hono/ci/mongodb-based-device-registry-values.yaml
kaniyan: Thanks for the info. I have updated it.

sophokles73 (approved):
LGTM

PR #116, opened 2020-07-07
Make tag name of Hono images configurable

sophokles73:
The chart has supported configuration of the container image name for
all components already. However, the tag name has been part of the image
name so far. In order to allow setting the version of all Hono
components in one go, the honoImagesTag property has been introduced
which defaults to the chart's appVersion. The property can be used for
example to use not yet released (milestone) images for experimenting
purposes. All of Hono's components also support a specific *imageTag*
property to set the image's tag name per component.

### DIFF
--- a/charts/hono/values.yaml
+++ b/charts/hono/values.yaml
@@ -14,6 +14,12 @@
 # Default values for eclipse-hono.
 # Declare variables to be passed into your templates.
 
+# honoImagesTag contains the (common) tag name of the Hono component
+# container images to deploy.
+# If not set explicitly, the chart's appVersion is used as the tag name.
+# Alternatively, the tag name can also be set per component image.
+# honoImagesTag:
+
 amqpMessagingNetworkExample:
   # enabled indicates whether the example AMQP Messaging Network
   # consisting of a single Dispatch Router and Broker should be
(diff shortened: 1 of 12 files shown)


charts/hono/values.yaml
```diff
@@ -328,7 +334,10 @@ adapters:
     enabled: true
     # imageName contains the name (including registry and tag)
     # of the container image to use for the AMQP adapter
-    imageName: index.docker.io/eclipse/hono-adapter-amqp-vertx:1.2.3
+    imageName: index.docker.io/eclipse/hono-adapter-amqp-vertx
+    # imageTag contains the tag of the container image to deploy.
+    # If not specified, the value of the honoImagesTag property is used.
+    # imageTag:
```
calohmn, line 340: According to what is used in the individual component yaml files and the README, this has to be named `tagName` instead of `imageTag`.

To verify:
````
helm install --dry-run --debug --dependency-update -n hono --set honoImagesTag=1.3.0-M3 --set adapters.amqp.tagName=1.2.3 eclipse-hono hono/ | grep "\- image:"
````
gives the expected results.

But
````
helm install --dry-run --debug --dependency-update -n hono --set honoImagesTag=1.3.0-M3 --set adapters.amqp.imageTag=1.2.3 eclipse-hono hono/ | grep "\- image:"
````
doesn't.


I would prefer the term `imageTag` here actually, which would mean changing individual component yaml files and the README accordingly.

sophokles73: thanks for spotting. I have no clue why it ended up this way in the deployment descriptors. It should definitely be *imageTag* (consistent with *imageName*). I'll fix that accordingly ...

sophokles73:
@calohmn I have pushed some corresponding changes ...

PR #130, opened 2020-08-07
Add Quarkus based images profiles and docs

dejanb:


dejanb:
@sophokles73 Yeah, that makes sense. This is just a quick way to test the current images.

I can mark this PR as draft and work on version with the configuration flag. How about we use something like

```imageType: spring```

with optional (for now) values of `quarkus` and `quarkus-native`?

**Comment by dejanb** [comment]
@sophokles73 Here's the version based on the `honoImagesType` property. Let me know how it looks to you.

File: charts/hono/templates/_helpers.tpl [hunk 1]
```diff
@@ -252,6 +252,21 @@ resource-limits:
...
 
+{{/*
+Configuration for the image name for services.
+The scope passed in is expected to be a dict with keys
+- "dot": the root scope (".") and
+- "conf": the configuration of the component
+*/}}
+{{- define "hono.imageName" -}}
+{{- $tag := ( default .dot.Chart.AppVersion ( default .dot.Values.honoImagesTag .dot.Values.adapters.http.imageTag ) ) -}}
+{{- if ( contains "quarkus" .dot.Values.honoImagesType ) }}
```
sophokles73, line 263: sure this is *contains*? looks like this branch is only applicable to JVM Quarkus images but not native images. or am I mistaken?

charts/hono/templates/_helpers.tpl
```diff
@@ -350,6 +366,18 @@ Optionally, the scope my contain key
...
+- name: "config"
+{{- if ( eq "quarkus" .dot.Values.honoImagesType ) }}
+  mountPath: "/deployments/config"
+{{- end }}
+{{- if ( eq "quarkus-native" .dot.Values.honoImagesType ) }}
+  mountPath: "/work/config"
+{{- end }}
+  readOnly: true
+{{- end }}
+{{- end }}
```
sophokles73, line 380: my understanding is that Quarkus expects to find the config files in different locations than `/etc/hono` and also cannot be configured to read them from `/etc/hono` (which I would find very strange), right?
If so, then I think that the solution should not be to duplicate the config volume but instead simply mount the config volume to the Quarkus specific path(s). This would simply require to add the if block as an *else if* block to the original volume mount, wouldn't it?

charts/hono/templates/_helpers.tpl
dejanb: By default with Quarkus you can build either JVM or native image and they will end up having the same name. So in that sense this will work with the current Hono implementation. We should discuss how we're gonna build and name images in Hono and adapt this according to that.

charts/hono/templates/_helpers.tpl
dejanb: The issue is only location of the `application.yaml` file. There are some more files in the secret that are referenced from the app config, so I tried to keep changes minimal. As you said, I see the issue of not having a configurable app config file as an issue we need to deal in Quarkus. In that sense I would not invest too much resources in this as it's only a temporary state.

charts/hono/templates/_helpers.tpl
sophokles73: I see

sophokles73 (commented):
IMHO we should also add new profiles to the `ci` folder for testing if deployment of the quarkus images works

dejanb:
> IMHO we should also add new profiles to the `ci` folder for testing if deployment of the quarkus images works

@sophokles73 Can we do that, considering that we're not publishing Quarkus images yet?

sophokles73:
> Can we do that, considering that we're not publishing Quarkus images yet?

@dejanb obviously not, my bad ;-)

sophokles73 (approved):
LGTM

PR #132, opened 2020-08-20
make service ports configurable

vhdirk:
We'd like to run this in 2 separate namespaces in a single cluster, so ports have to be configurable.

Also; all of these services are configured as either LoadBalancer or NodePort. Why is that? It would be far easier if this would be either LoadBalancer or ClusterIp, for http at least.

Any ideas @ctron @dejanb? 

Signed-off-by: Dirk Van Haerenborgh <dirk.vanhaerenborgh@aloxy.io>

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.4.3
+version: 1.4.4
 # Version of Hono being deployed by the chart
 appVersion: 1.4.0
 keywords:
--- a/charts/hono/templates/_helpers.tpl
+++ b/charts/hono/templates/_helpers.tpl
@@ -415,3 +415,29 @@ Optionally, the scope my contain key
 {{- end }}
 {{- end }}
 {{- end }}
+
+{{/*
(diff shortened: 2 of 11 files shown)


sophokles73:
@vhdirk you'll also need to bump the chart's version (to `1.3.16`) in order for the CI pipeline to succeed.

sophokles73:
@vhdirk would you mind rebasing and increasing the chart version so that the CI pipeline kicks in?

vhdirk:
sure, no problem :)

sophokles73 (approved):
LGTM

sophokles73:
@ctron would you mind taking another look? FMPOV this can be merged ...

ctron (approved):

charts/hono/Chart.yaml
```diff
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.4.2
+version: 1.4.3
```
sophokles73, line 18: the chart version needs to be bumped once more because there have been other changes in the meantime. Once this is done, we can merge the PR.

sophokles73 (changes requested):
One final detail ...
sorry for the delay and thanks for your patience :+1:

charts/hono/Chart.yaml
vhdirk: done!

PR #134, opened 2020-09-01
updated cloud2edge package to latest Ditto+Hono

thjaeckle:
* use Hono 1.3.0
* use Ditto 1.2.0
* configure Ditto/Hono connection template to:
   * configure Ditto "acknowledgementRequests" to activate QoS 1 processing for Hono messages with "qos" header set to "1"
   * configure a "replyTarget" for consumed telemetry/event messages from Hono - when the Ditto Protocol payload message contains a "reply-to" header, a response is sent back to that address

I thought using the newest minor versions of Hono/Ditto is worth a minor update of the c2e package, so I chose version `0.1.0`.
What do you think?

### DIFF
--- a/packages/cloud2edge/Chart.yaml
+++ b/packages/cloud2edge/Chart.yaml
@@ -11,8 +11,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.0.10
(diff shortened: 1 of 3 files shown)


sophokles73 (approved):
LGTM

PR #140, opened 2020-09-21
[#136] Use Hono chart with new demo keys/certs in c2e

calohmn:
Upgrade Hono dependency in Cloud2Edge package to the one with the recreated demo keys/certs (#136).


sophokles73 (approved):
LGTM

PR #141, opened 2020-09-21
Upgrade to hono 1.4.0

sophokles73:


charts/hono/templates/hono-adapter-amqp/headless-svc.yaml
```diff
@@ -1,16 +0,0 @@
-{{- if and .Values.adapters.amqp.enabled .Values.prometheus.createInstance }}
```
calohmn, line 1: The headless services should then also be removed from the `hono/README.md` file.

Btw., what is the remaining `eclipse-hono-service-device-registry-headless` service used for?

sophokles73: exactly

sophokles73:
@calohmn done ...

calohmn (approved):
LGTM

Issue #142, opened 2020-09-22
Error creating cloud2edge demo device in Hono


calohmn:
As reported in the Hono Gitter channel:
When using the cloud2edge package (chart version > 0.0.10), creating the device with the identifier "org.eclipse.packages.c2e:device" (containing a colon, as required by Ditto) fails with the error:
````
failed validation: value does not match pattern [^[a-zA-Z0-9-.]++$]
````

After configuring a working `deviceIdPattern` to fix this, there is an error updating the credentials:
````
ClientErrorException: request body does not contain valid Credentials objects
````

sophokles73:
> After configuring a working deviceIdPattern to fix this, there is an error updating the credentials:
`ClientErrorException: request body does not contain valid Credentials objects`

@calohmn 
If setting another pattern does not fix the problem, then why do you want to address this issue by means of setting another pattern?

calohmn:
@sophokles73 Setting the pattern in the Hono chart is only the first part of the fix.
The 2nd part of the fix is using the fixed Hono chart in the cloud2edge package and adapting the "demo-device-credentials.json" file there (removing "device-id" entries which lead to the credentials parsing error).

sophokles73:
Why would any of the device IDs cause any trouble after the pattern has been fixed to allow for the `:` character?

calohmn:
@sophokles73 This second issue concerning the `request body does not contain valid Credentials objects` error has nothing to do with the device ID pattern check.
The "demo-device-credentials.json" contains "device-id" entries which aren't specified in the credentials schema. In Hono 1.2, these entries were ignored, in later versions these cause an error (`@JsonIgnoreProperties(ignoreUnknown = true)` was removed in `CommonCredential`).

sophokles73:
Ah, now I see. Thanks @calohmn :+1:

PR #143, opened 2020-09-22
[#142] Use deviceIdPattern with added colon

calohmn:
This is for #142:
Allow Hono device identifiers to contain a colon, as required by Ditto.
Also fix some typos.

### DIFF
--- a/charts/hono/templates/hono-service-device-registry/hono-service-device-registry-secret.yaml
+++ b/charts/hono/templates/hono-service-device-registry/hono-service-device-registry-secret.yaml
@@ -63,6 +63,7 @@ stringData:
           insecurePortEnabled: true
           insecurePortBindAddress: "0.0.0.0"
           {{- end }}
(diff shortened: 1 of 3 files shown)


charts/hono/templates/hono-service-device-registry/hono-service-device-registry-secret.yaml
```diff
@@ -63,6 +63,7 @@ stringData:
           insecurePortEnabled: true
           insecurePortBindAddress: "0.0.0.0"
           {{- end }}
+          deviceIdPattern: "^[a-zA-Z0-9-_\\.\\:]++$"
```
sophokles73, line 66: What's the purpose of the second `+` before the `$`?

calohmn: Good question, none, I guess :). Didn't pay attention here when adopting it from Hono, where it's also superfluous.

sophokles73 (approved):
LGTM

PR #144, opened 2020-09-23
[#142] Use updated Hono chart; fix credentials json

### LINKED ISSUE #142: Error creating cloud2edge demo device in Hono

calohmn:
This fixes #142.
Use updated Hono chart with fix to allow device identifiers to contain a colon, as required by Ditto.
Remove invalid "device-id" fields in credentials json.

### DIFF
--- a/packages/cloud2edge/Chart.yaml
+++ b/packages/cloud2edge/Chart.yaml
@@ -11,8 +11,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.1.1
-appVersion: 0.1.1
+version: 0.1.2
+appVersion: 0.1.2
 name: cloud2edge
 description: |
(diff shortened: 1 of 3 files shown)


calohmn:
This includes an update from Hono 1.3 to Hono 1.4 in the cloud2edge package.
I have updated only the package chart *patch* version here since the corresponding Hono 1.3->1.4 chart update also only updated the *patch* version, there being no incompatibilites between the versions.

sophokles73 (approved):
LGTM

PR #147, opened 2020-09-25
Avoid config property name clash

sophokles73:
The default name used by the MongoDB chart for creating the
corresponding Kubernetes Service will cause a name clash with the
"HONO_MONGO_DB" configuration variable of the MongoDB based registry
when using "hono" as the Helm release name during installation.

Overriding the chart's name with "mongodb-server" prevents this from
happening.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.4.1
+version: 1.4.2
 # Version of Hono being deployed by the chart
 appVersion: 1.4.0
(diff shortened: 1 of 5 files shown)


charts/hono/Chart.yaml
```diff
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.4.0
+version: 1.4.1
```
calohmn, line 18: Can you increase the version here? The version has already been updated to 1.4.1 in the meantime.

sophokles73: shoot

sophokles73: will do

calohmn (approved):
LGTM

PR #149, opened 2020-09-30
[hawkbit] update ingress version for kubernetes >= 1.14

rpahli:
update ingress version to `networking.k8s.io/v1beta1` for kubernetes version >= 1.14

### DIFF
--- a/charts/hawkbit/templates/ingress.yaml
+++ b/charts/hawkbit/templates/ingress.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "hawkbit.fullname" . -}}
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
   name: {{ $fullName }}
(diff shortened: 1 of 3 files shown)


charts/hawkbit/templates/ingress.yaml
```diff
@@ -1,6 +1,10 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "hawkbit.fullname" . -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
```
monotek, line 3: i would remove this because helm 3 likely does not work with it at the moment.
see: https://github.com/helm/helm/issues/7004

monotek (changes requested):

monotek (changes requested):
The READEM.md says compatible from k8s 1.11.
Should be changed too.

rpahli:
fixed

monotek (approved):

laverman:
Thanks for your contribution. However, I'm going to close this PR in favor of #150

Issue #155, opened 2020-10-29
Cannot mount secret into the Artemis container's filestystem


damian-gallo:
According to the `values.yaml` file the mounted secrets should be specified at `hono.amqpMessagingNetworkExample.broker.artemis.extraSecretMounts` but only works when specified at `hono.amqpMessagingNetworkExample.broker.extraSecretMounts` (without "artemis").

It seems that the `artemis-deployment.yaml` is pointing at the wrong value:
`.Values.amqpMessagingNetworkExample.broker`
where it should be:
`.Values.amqpMessagingNetworkExample.broker.artemis`

sophokles73:
@kaniyan Hi Karthees, is this something for you?

kaniyan:
> @kaniyan Hi Karthees, is this something for you?

I will take a look at it.

sophokles73:
Ok, thanks :+1:

kaniyan:
@damian-gallo This issue has been fixed via PR #158 and this fix is there from version 1.4.5. Could you please take a look and close this issue?

damian-gallo:
Thank you very much!

PR #158, opened 2020-11-02
[#155] Fix mounting of additional secrets into the Artemis container's filesystem

kaniyan:
With this fix, now it should be able to mount the additional secrets specified in `values.yaml` into the Artemis container's file system.


calohmn (approved):
LGTM

PR #159, opened 2020-11-06
Update to Hono 1.4.3

sophokles73:


charts/hono/requirements.yaml
```diff
@@ -13,11 +13,11 @@
 dependencies:
   - name: prometheus
     version: ~11.7.0
-    repository: "https://kubernetes-charts.storage.googleapis.com/"
+    repository: "https://charts.helm.sh/stable"
```
calohmn, line 16: Starting with the newer chart version 11.12.1, the Prometheus chart is marked as deprecated in the helm stable repository (see [here](https://github.com/helm/charts/tree/master/stable/prometheus)).

The repo with the newest chart version is `https://prometheus-community.github.io/helm-charts`, so I think we could use that repository here now already (version 11.7.0 is in there as well).

charts/hono/requirements.yaml
```diff
@@ -13,11 +13,11 @@
...
     condition: prometheus.createInstance
   - name: grafana
     version: ~3.8.15
-    repository: "https://kubernetes-charts.storage.googleapis.com/"
+    repository: "https://charts.helm.sh/stable"
```
calohmn, line 20: The Grafana chart has moved as well (see [here](https://github.com/helm/charts/tree/master/stable/grafana)).

So we could use `https://grafana.github.io/helm-charts` here instead already.

ctron (approved):

calohmn (approved):

PR #161, opened 2020-11-12
allow users to turn off example data in the device registry

BobClaerhout:


ctron (commented):
Looks like there are some spaces for the linter to complain about: :+1: in any case!

ctron (approved):

PR #162, opened 2020-11-16
Use azure file shares as persistent storage

superkartoffel:
Hi,

I would like to use azure file shares as persistent storage because they are easier to scale in my case. In this PR I have added two options:

with fileStorage.azure.dynamicFileShare kubernetes will create a file share for you.
with fileStorage.azure.staticSecret you can specify a file share that you have created manually.


ctron:
Thank you @superkartoffel for the PR!

I personally would prefer a more generic solution to this. Adding every custom storage provider to the helm chart won't scale.

Maybe you can take some inspiration from the Bitnami Helm charts on how to achieve this.

Also think it is wrong to crate a `StorageClass` instance in the hawkBit helm chart. As this is a cluster wide resource, I don't think it is appropriate to let an application helm chart create this.

PR #169, opened 2020-11-30
[#168] Use ANYCAST routing type in broker config

### LINKED ISSUE #168: [Hono] Event messages aren't queued when there is no consumer
Sending `event` messages to a Hono protocol adapter should result in the messages getting queued in the broker when there is no northbound consumer for the messages currently connected.

But that doesn't seem to be the case. `event` messages only get delivered to a northbound consumer application when the application is connected while the message is being sent.

calohmn:
This fixes #168:
This fixes the issue of queues not getting created when event messages are sent and of event messages getting forwarded to multiple consumers.


sophokles73 (approved):
LGTM

Issue #176, opened 2020-12-17
Ditto chart can no longer be packaged by Helm in website build job


sophokles73:
The chart repository's website build job seems to be failing [1] while trying to package the Ditto chart.

[1] https://ci.eclipse.org/packages/job/Website/job/master/84/console

sophokles73:
This might be due to the deprecation [1] of the https://kubernetes-charts.storage.googleapis.com chart repo

[1] https://github.com/helm/charts/

PR #177, opened 2020-12-17
[#176] Use bitnami chart repo URL in Ditto's requirements

sophokles73:


charts/ditto/requirements.yaml
```diff
@@ -11,6 +11,6 @@
 ---
 dependencies:
   - name: mongodb
-    repository: https://kubernetes-charts.storage.googleapis.com
+    repository: https://charts.bitnami.com/bitnami
     version: 7.6.3
```
thjaeckle, line 15:
```suggestion
    version: ~7.14.7
```
Should we use the same version as Hono defines?

thjaeckle (changes requested):
Thank you for fixing this, @sophokles73 
The deprecation seems indeed to be the issue here.

I think it makes sense to define the same MongoDB version as Hono defines in order to be sure that a working MongoDB version is referenced ;)

charts/ditto/requirements.yaml
sophokles73: Well, I am glad to align the versions. However, you will need to tell me if Ditto will work with it ...

thjaeckle: Yes, it will - this references the latest MongoDB 4.2 which Ditto 1.5.0 was explicitly tested against.

sophokles73: done

thjaeckle (approved):

PR #182, opened 2020-12-23
Fix repo URLs in GitHub Actions workflow

calohmn:
Replacing the obsolete "https://kubernetes-charts.storage.googleapis.com"
location and adding prometheus and grafana repos.

PR #186, opened 2021-01-15
Update to latest kubernetes version(s)

sophokles73:
The chart linter and kind deploy confiugrations have been updated to the
latest kubernetes versions.
The helm/kind-action has been upgraded to the latest minor version
(1.1.0).

### DIFF
--- a/.github/kind-config.yaml
+++ b/.github/kind-config.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2019 Contributors to the Eclipse Foundation
+# Copyright (c) 2019, 2021 Contributors to the Eclipse Foundation
 #
 # See the NOTICE file(s) distributed with this work for additional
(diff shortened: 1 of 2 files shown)


calohmn:
I think it would be good to add a temporary test commit in this PR, updating the versions in the different IoT Packages charts, so that the GitHub actions workflow is triggered.
I remember getting errors concerning certain K8s resources (I think from hawkbit) when trying to update "chart-testing-action" to v1.1.0, which seemed to cause a newer K8s version to get used (in #182).

I guess also the apiVersion in ".github/kind-config.yaml" might need to be updated from the current "apiVersion: kind.sigs.k8s.io/v1alpha3".

sophokles73:
@calohmn WDYT?

calohmn:
@sophokles73 looking good! Chart changes can be reverted again.

PR #189, opened 2021-01-18
add 'subPath' support for secretVolumeMounts in Hono chart

eriksven:
This PR extends the 'defineSecretVolumeMounts' in the _helpers.tpl used within the Hono chart to also consider the 'subPath' property.

Note, since the previous PR #187 increases the version field to 1.4.22, this PR already sets the version to 1.4.23. Therefore, I propose to merge the other PR (#187) before this PR.


sophokles73 (commented):
looks good
can you rebase and resolve the conflict?

sophokles73 (approved):
LGTM

PR #191, opened 2021-01-21
Add messaging config for device registries

calohmn:
The gateway-based auto-provisioning feature requires that the device registry connects to the AMQP messaging network. The necessary configuration is added here.

This is a preparation for Hono 1.6.


sophokles73 (approved):
LGTM

PR #193, opened 2021-02-03
fixed nginx CORS config for Ditto chart

thjaeckle:
* for some browsers, usage of "*" for "Access-Control-Allow-Origin" is forbidden - fix that
* add "PATCH" verb
* white-list all HTTP headers


ctron (approved):

stmaute:
LGTM 👍🏻 
Already had prepared the same changes.

stmaute (approved):

PR #197, opened 2021-02-05
allow specifying custom certificates for adapters

### LINKED ISSUE #196: Hono: Use adapter custom certificates in registry service clients.
It's currently not possible to use the `adapters.tenantSpec`, `adapters.deviceRegistrationSpec`, `adapters.credentialsSpec`, `adapters.deviceConnectionSpec` properties of `values.yaml` to configure the registry service clients.
Those are ignore when `deviceRegistryExample.enabled=false`, but then you're not deploying a device registry to begin with.

vhdirk:
Fixes #196

### DIFF
--- a/charts/hono/templates/_helpers.tpl
+++ b/charts/hono/templates/_helpers.tpl
@@ -207,9 +207,7 @@ command:
 {{- end }}
 tenant:
 {{- if .dot.Values.adapters.tenantSpec }}
-  {{- range $key, $value := .dot.Values.adapters.tenantSpec }}
-  {{ $key }}: {{ $value }}
-  {{- end }}
+  {{- .dot.Values.adapters.tenantSpec | toYaml | nindent 2 }}
 {{- else if .dot.Values.deviceRegistryExample.enabled }}
   name: Hono {{ $adapter }}
   host: {{ .dot.Release.Name }}-service-device-registry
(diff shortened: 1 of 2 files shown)


charts/hono/templates/_helpers.tpl
```diff
@@ -206,7 +206,11 @@ command:
   {{- required ".Values.adapters.commandAndControlSpec MUST be set if example AMQP Messaging Network is disabled" .dot.Values.adapters.commandAndControlSpec | toYaml | nindent 2 }}
 {{- end }}
 tenant:
-{{- if .dot.Values.deviceRegistryExample.enabled }}
+{{- if .dot.Values.adapters.tenantSpec }}
+  {{- range $key, $value := .dot.Values.adapters.tenantSpec }}
+  {{ $key }}: {{ $value }}
+  {{- end }}
```
sophokles73, line 212: how about simply using
```
{{- .dot.Values.adapters.tenantSpec | toYaml | nindent 2 }}
```

vhdirk: That's cleaner indeed. Also replaced other occurances.

sophokles73 (approved):
LGTM

sophokles73:
Thanks for contributing :+1:

PR #199, opened 2021-02-16
[#198] Fix 'command-router' cache not being defined

### LINKED ISSUE #198: Command Router never becomes ready
When installing the Hono chart with `useCommandRouter=true` the Command Router pod never becomes ready. The log states the following:
```
09:15:06.216 [HotRod-client-async-pool-1-2] WARN  org.infinispan.HOTROD - ISPN004005: Error received from the server: org.infinispan.server.hotrod.CacheNotFoundException: Cache with name 'command-router' not found amongst the configured caches
09:15:16.054 [HotRod-client-async-pool-1-2] WARN  org.infinispan.HOTROD - ISPN004005: Error received from the server: org.infinispan.server.hotrod.CacheNotFoundException: Cache with name 'command-router' not found amongst the configured caches
09:15:26.046 [HotRod-client-async-pool-1-2] WARN  org.infinispan.HOTROD - ISPN004005: Error received from the server: org.infinispan.server.hotrod.CacheNotFoundException: Cache with name 'command-router' not found amongst the configured caches
09:15:36.045 [HotRod-client-async-pool-1-2] WARN  org.infinispan.HOTROD - ISPN004005: Error received from the server: org.infinispan.server.hotrod.CacheNotFoundException: Cache with name 'command-router' not found amongst the configured caches
09:15:46.061 [HotRod-client-async-pool-1-2] WARN  org.infinispan.HOTROD - ISPN004005: Error received from the server: org.infinispan.server.hotrod.CacheNotFoundException: Cache with name 'command-router' not found amongst the configured caches
09:15:56.024 [HotRod-client-async-pool-1-2] WARN  org.infinispan.HOTROD - ISPN004005: Error received from the server: org.infinispan.server.hotrod.CacheNotFoundException: Cache with name 'command-router' not found amongst the configured caches
```

It looks like the Command Router service expects to find a cache called `command-router` managed by the data grid. Taking a closer look at the example data grid's (Infinispan) configuration file, it indeed looks like the cache that the grid manages is named `device-connection`.

calohmn:
This fixes #198.

### DIFF
--- a/charts/hono/config/infinispan/hono-data-grid.xml
+++ b/charts/hono/config/infinispan/hono-data-grid.xml
@@ -14,7 +14,7 @@
         <role name="adapter" permissions="READ WRITE" />
       </authorization>
     </security>
-    <distributed-cache name="device-connection">
+    <distributed-cache name="{{ if .Values.useCommandRouter }}command-router{{ else }}device-connection{{ end }}">
       <encoding>
         <key media-type="text/plain" />
         <value media-type="text/plain" />
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.5.2
+version: 1.5.3
 # Version of Hono being deployed by the chart
 appVersion: 1.6.0
 keywords:


charts/hono/config/infinispan/hono-data-grid.xml
```diff
@@ -24,6 +24,16 @@
...
+    <distributed-cache name="command-router">
+      <encoding>
+        <key media-type="text/plain" />
+        <value media-type="text/plain" />
+      </encoding>
+      <memory storage="OFF_HEAP" max-size="30 MB" when-full="REMOVE"/>
+      <security>
+        <authorization enabled="true" />
+      </security>
+    </distributed-cache>
```
sophokles73, line 36: why do we need to maintain two caches? Don't they serve the exact same purpose?

calohmn: They do server the same purpose (for now). Nonetheless, I would rather use the default cache names. Otherwise the cache name would have to be set explicitly in the config, which would make things rather confusing.

But I've amended the commit to now only have one "distributed-cache" block, with either the "command-router" name or "device-connection".

sophokles73 (approved):
LGTM

Issue #152, opened 2020-10-05
Support using same Mongo DB instance


sophokles73:
The c2e package should support having Ditto and Hono use the same Mongo DB instance. The chart should support either installing a single node Mongo DB instance or configuring an already existing DB instance which will then be used by Hono and/or Ditto.

sophokles73:
The package actually already supports configuring Hono and Ditto to use an already existing Mongo DB instance. Hono can also be configured to use the instance started by Ditto when using the Mongo DB based registry. In fact, I wonder if we should make that the default setup, i.e. using the Mongo DB based registry and configure it to use Ditto's mongo instance.

WDYT? @ctron @thjaeckle @calohmn

thjaeckle:
That makes a lot of sense to me - the MongoDB is already in the c2e package, so why not use it also for the Hono device registry ..

sophokles73:
@calohmn @ctron any opinion?

calohmn:
Yes, I think that should be the default setup.

PR #203, opened 2021-03-09
[#202] Enable Hono Device Connection service

### LINKED ISSUE #202: Connection Target Error
https://github.com/eclipse/ditto/issues/971

Ditto can't publish command any more as I updated packages from 0.1.2 to 0.1.3. Is it because of COMMAND ROUTER which is updated recently in HONO??
Please help me out! Thanks

calohmn:
This fixes #202:
With the switch to the Hono MongoDB-based device registry (not containing a device connection service implementation like the file-based one does) in the cloud2edge chart (#152), there is no device connection service implementation getting deployed anymore, leading to Hono command & control errors. 

Therefore the "deviceConnectionService.enabled" option is set here to deploy the device connection service component, using an embedded cache (newer Hono chart version needed for that).

### DIFF
--- a/packages/cloud2edge/Chart.yaml
+++ b/packages/cloud2edge/Chart.yaml
@@ -11,8 +11,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.1.3
-appVersion: 0.1.3
+version: 0.1.4
+appVersion: 0.1.4
 name: cloud2edge
 description: |
   Eclipse IoT Cloud2Edge (C2E) is an integrated suite of services developers can use to build IoT applications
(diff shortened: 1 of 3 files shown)


sophokles73:
@calohmn is there a particular reason why you want to use the Device Connection service instead of the Command Router?

calohmn:
Usage of the Command Router isn't the default in the Hono chart yet. 
I would like to migrate to using the Command Router per default in the Hono and cloud2edge charts when updating to Hono 1.7, having removed the "tech preview" status there (see https://github.com/eclipse/hono/issues/2535).

sophokles73 (approved):
LGTM

PR #208, opened 2021-03-12
fixed device registry http route wrongly pointing to amqps endpoint

daFritz84:
The http device registry route was pointing to the `amqps` endpoint, which should point to the `http` endpoint. 
Tested on okd 4.7 with `curl`.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.5.6
+version: 1.5.7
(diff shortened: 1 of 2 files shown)


calohmn:
@daFritz84 Thanks for the PR. For it to be accepted, you first need to sign the Eclipse Contributor Agreement and then set the Signed-off-by field with your email address in the commit (you can amend and then force-push the commit), as described [here](https://github.com/eclipse/packages/blob/master/CONTRIBUTING.md#eclipse-contributor-agreement). You will also need to increase the Hono chart version here:
https://github.com/eclipse/packages/blob/b9039398c5a04172d6082d8053a7b3ba91bc7d83/charts/hono/Chart.yaml#L18
(Increase the patch version to 1.5.7.)
Having done so, the tests should be green.

calohmn (approved):
LGTM. Thanks for fixing this.

PR #215, opened 2021-04-07
Hono: Adapt setting of 'Always' pull policy

calohmn:
An "Always" "imagePullPolicy" will only be used on SNAPSHOT images if the image name doesn't
start with "index.docker.io/eclipse/". 
This simplifies usage of self-built images with standard names in minikube. And the "index.docker.io" registry doesn't contain "eclipse/*" SNAPSHOT images anyway.


sophokles73 (approved):
LGTM

PR #221, opened 2021-04-27
Update to Hono 1.7.1, fix example-tenants.sh

calohmn:
Updates the Hono chart to Hono 1.7.1 and fixes the wrong DEFAULT_TENANT trusted-ca data in `example.tenants.sh` (see https://github.com/eclipse/hono/issues/2628).


charts/hono/example/create_certs.sh
```diff
@@ -126,6 +126,14 @@ echo "trusted-ca.algorithm=$KEY_ALG" >> $DIR/trust-anchor.properties
 echo "trusted-ca.not-before=$NOT_BEFORE" >> $DIR/trust-anchor.properties
 echo "trusted-ca.not-after=$NOT_AFTER" >> $DIR/trust-anchor.properties
 
+echo "{" > default_tenant-trusted-ca.json
```
sophokles73, line 129: the lines above where the variables are added to the `trust-anchor.properties` file seem to be not needed at all, or are they? If not, then I suggest to remove them ...

sophokles73 (approved):
LGTM

PR #229, opened 2021-05-06
Prepare for Ditto 2.0

stmaute:
Update to Ditto 2.0 and update deployment files.


charts/ditto/Chart.yaml
```diff
@@ -14,10 +14,9 @@ name: ditto
 description: |
   Eclipse Ditto™ is a technology in the IoT implementing a software pattern called “digital twins”.
   A digital twin is a virtual, cloud based, representation of his real world counterpart
-  (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids,
-  EV charging stations etc).
-version: 1.5.3
-appVersion: 1.5.1
+  (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
+version: 1.5.4
```
thjaeckle, line 18: I would rather have the `version` be in sync with the `appVersion`, so please adjust to `2.0.0` here as well.

charts/ditto/Chart.yaml
```diff
@@ -31,3 +30,5 @@ maintainers:
   email: axel.koehler@kiwigrid.com
 - name: thjaeckle
   email: thomas.jaeckle@bosch-si.com
+- name: stmaute
+  email: stefan.maute@bosch-si.com
```
thjaeckle, line 34: newline is missing from file - that's what the Helm linter also complains about

charts/ditto/Chart.yaml
stmaute: that is a good idea to keep this in sync 👍🏻

charts/ditto/Chart.yaml
stmaute: already fix it

thjaeckle (approved):
LGTM 👍

PR #232, opened 2021-05-12
Use camel case for resource limit checks configuration

### LINKED ISSUE #230: [Hono] Prometheus based resource limit check can't be used with native images
The native image variant of the protocol adapters fail to pick up the Prometheus server configuration during start up, thus falling back to the noop resource limit checks.

sophokles73:
The Quarkus based native adapter images require the configuration
properties in application.yml to be all camel case.

Fixes #230


calohmn (approved):
LGTM

PR #237, opened 2021-05-18
[#222] Support Kafka based messaging in Hono.

### LINKED ISSUE #222: Support Kafka-based messaging in Hono
With the ongoing efforts in the Hono project to support Kafka as a messaging system instead of AMQP 1.0, the packages project should add support for this feature. 
To support an easy start, it should be possible to deploy a Kafka broker with Hono directly.

b-abel:
This fixes #222. It adds the support for Kafka-based messaging to the Hono chart. The property `messagingNetworkType` is used to select which type of configuration is added to the adapters and services (either `amqp` or `kafka`). The property `kafkaMessagingClusterExample.enabled` indicates whether the example Kafka cluster should be deployed and used.


b-abel:
## How to use this

### Deploy to minikube

First, build Hono images with the version `1.8.0-SNAPSHOT`.
Then use this command from the current branch to deploy:

```bash
# in directory packages/charts/hono
helm install --dependency-update -n hono --values ./profileKafkaMessaging.yaml --set honoImagesTag=1.8.0-SNAPSHOT eclipse-hono .
```

### Send events

Start receiver in the Hono project:

```sh
# in directory: hono/cli/target/
export MY_TENANT=DEFAULT_TENANT
export MY_DEVICE=4711
export KAFKA_IP=$(kubectl get service eclipse-hono-kafka-0-external --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
java -jar hono-cli-*-exec.jar --hono.client.username=consumer@HONO --hono.client.password=verysecret --tenant.id=$MY_TENANT --device.id=$MDEVICE --spring.profiles.active=receiver,kafka --hono.kafka.commonClientConfig.bootstrap.servers=$KAFKA_IP:9094
```

Send the event:

```sh
export MY_TENANT=DEFAULT_TENANT
export MY_DEVICE=sensor1
export MY_PWD=hono-secret
export HTTP_ADAPTER_IP=$(kubectl get service eclipse-hono-adapter-http-vertx --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
curl -i -u $MY_DEVICE@$MY_TENANT:$MY_PWD -H 'Content-Type: application/json' --data-binary '{"alarm": "fire"}' http://$HTTP_ADAPTER_IP:8080/event
```

### Send and receive Commands

Start a simulated devices:

```sh
export MY_TENANT=DEFAULT_TENANT
export MY_DEVICE=sensor1
export MY_PWD=hono-secret
export MQTT_ADAPTER_IP=$(kubectl get service eclipse-hono-adapter-mqtt-vertx --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
mosquitto_sub -v -h $MQTT_ADAPTER_IP -u $MY_DEVICE@$MY_TENANT -P $MY_PWD -t command/+/+/req/#
```

Send a command with Hono's CLI from the Hono project:

```sh
# in directory: hono/cli/target/
export MY_TENANT=DEFAULT_TENANT
export MY_DEVICE=4711
export KAFKA_IP=$(kubectl get service eclipse-hono-kafka-0-external --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
java -jar hono-cli-*-exec.jar --hono.client.username=consumer@HONO --hono.client.password=verysecret --tenant.id=$MY_TENANT --device.id=$MY_DEVICE --spring.profiles.active=command,kafka --hono.kafka.commonClientConfig.bootstrap.servers=$KAFKA_IP:9094
```

b-abel:
The [chart](https://github.com/bitnami/charts/tree/master/bitnami/kafka) used for the deployment of Kafka requires Helm version 3.1.0. With a lower version, the deployment might not work when the property `kafkaMessagingClusterExample.enabled` is set to `true`.

b-abel:
I assume the currently failing lint check is caused by #239. This should be resolved with #240.

calohmn:
@b-abel Can you rebase and update the chart version?

b-abel:
Rebased.

b-abel:
## How to use Kafka-based Messaging with Authentication and TLS

### Deploy to minikube

Deploy from the current branch:

```bash
# in directory packages/charts/hono
helm install --dependency-update -n hono --values ./profileKafkaMessaging.yaml --set honoImagesTag=1.8.0 eclipse-hono .
```

### Send events

Download Hono's [Command Line Client](https://www.eclipse.org/downloads/download.php?file=/hono/hono-cli-1.8.0-exec.jar).
Change the path to the packages project in the following script block and run it to start a receiver:

```sh
# in directory where the hono-cli-*-exec.jar file has been downloaded to
export PATH_TO_PACKAGES=<path/to/the/packages/project>
export MY_TENANT=DEFAULT_TENANT
export MY_DEVICE=4711
export KAFKA_IP=$(kubectl get service eclipse-hono-kafka-0-external --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
java -jar hono-cli-*-exec.jar --tenant.id=$MY_TENANT --device.id=$MY_DEVICE --spring.profiles.active=receiver,kafka --hono.kafka.commonClientConfig.bootstrap.servers=$KAFKA_IP:9094 --hono.kafka.commonClientConfig.security.protocol=SASL_SSL --hono.kafka.commonClientConfig.sasl.mechanism=SCRAM-SHA-512 --hono.kafka.commonClientConfig.sasl.jaas.config="org.apache.kafka.common.security.scram.ScramLoginModule required username=\"hono\" password=\"hono-secret\";" --hono.kafka.commonClientConfig.ssl.endpoint.identification.algorithm="" --hono.kafka.commonClientConfig.ssl.truststore.location=$PATH_TO_PACKAGES/charts/hono/example/certs/trustStore.jks --hono.kafka.commonClientConfig.ssl.truststore.password=honotrust
```

Send the event:

```sh
export MY_TENANT=DEFAULT_TENANT
export MY_DEVICE=sensor1
export MY_PWD=hono-secret
export HTTP_ADAPTER_IP=$(kubectl get service eclipse-hono-adapter-http-vertx --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
curl -i -u $MY_DEVICE@$MY_TENANT:$MY_PWD -H 'Content-Type: application/json' --data-binary '{"alarm": "fire"}' http://$HTTP_ADAPTER_IP:8080/event
```

### Send and receive Commands

Start a simulated devices:

```sh
export MY_TENANT=DEFAULT_TENANT
export MY_DEVICE=sensor1
export MY_PWD=hono-secret
export MQTT_ADAPTER_IP=$(kubectl get service eclipse-hono-adapter-mqtt-vertx --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
mosquitto_sub -v -h $MQTT_ADAPTER_IP -u $MY_DEVICE@$MY_TENANT -P $MY_PWD -t command/+/+/req/#
```

Change the path to the packages project in the following script block and run it to send a command with Hono's CLI:

```sh
# in directory where the hono-cli-*-exec.jar file has been downloaded to
export PATH_TO_PACKAGES=<path/to/the/packages/project>
export MY_TENANT=DEFAULT_TENANT
export MY_DEVICE=4711
export KAFKA_IP=$(kubectl get service eclipse-hono-kafka-0-external --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
java -jar hono-cli-*-exec.jar --tenant.id=$MY_TENANT --device.id=$MY_DEVICE --spring.profiles.active=command,kafka --hono.kafka.commonClientConfig.bootstrap.servers=$KAFKA_IP:9094 --hono.kafka.commonClientConfig.security.protocol=SASL_SSL --hono.kafka.commonClientConfig.sasl.mechanism=SCRAM-SHA-512 --hono.kafka.commonClientConfig.sasl.jaas.config="org.apache.kafka.common.security.scram.ScramLoginModule required username=\"hono\" password=\"hono-secret\";" --hono.kafka.commonClientConfig.ssl.endpoint.identification.algorithm="" --hono.kafka.commonClientConfig.ssl.truststore.location=$PATH_TO_PACKAGES/charts/hono/example/certs/trustStore.jks --hono.kafka.commonClientConfig.ssl.truststore.password=honotrust
```

### Connect external tools to the Kafka cluster

The following example shows how the Kafka CLI can be used as a northbound receiver.

First, [download Kafka](https://www.apache.org/dyn/closer.cgi?path=/kafka/2.8.0/kafka_2.13-2.8.0.tgz) and extract it like this:

```sh
tar -xzf kafka_2.13-2.8.0.tgz
cd kafka_2.13-2.8.0
```

Change the path to the packages project in the following script block and run it to create a client config file:

```sh
export PATH_TO_PACKAGES=<path/to/the/packages/project>
cat <<EOS >> hono-client-sasl-ssl.config
security.protocol=SASL_SSL
sasl.mechanism=SCRAM-SHA-512
sasl.jaas.config=org.apache.kafka.common.security.scram.ScramLoginModule required \
  username="hono" \
  password="hono-secret";
# disable hostname verification because the certificate does not contain load balancer IPs
ssl.endpoint.identification.algorithm=
ssl.truststore.location=$PATH_TO_PACKAGES/charts/hono/example/certs/trustStore.jks
ssl.truststore.password=honotrust
EOS
```

Now, start the console consumer on the _events_ topic of the tenant:

```sh
export MY_TENANT=DEFAULT_TENANT
export MY_DEVICE=4711
export TOPIC=hono.event.$MY_TENANT
export KAFKA_IP=$(kubectl get service eclipse-hono-kafka-0-external --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
bin/kafka-console-consumer.sh --topic $TOPIC --bootstrap-server $KAFKA_IP:9094 --consumer.config ./hono-client-sasl-ssl.config
```


To list all topics in Kafka, run:

```sh
bin/kafka-topics.sh --bootstrap-server $KAFKA_IP:9094 --list --command-config ./hono-client-sasl-ssl.config
```

b-abel:
Regarding the documentation I think it would be best to have the "How to..." part from [above](https://github.com/eclipse/packages/pull/237#issuecomment-847950264) integrated somehow into Hono's [Getting Started](https://www.eclipse.org/hono/getting-started/) instead of adding more documentation here.

b-abel:
This adds When deploying the dispatch router, there is one port for TLS and one for plaintext connections. The Kafka cluster is now configured so that clients can only connect with TLS. At least with Bitnami's Helm chart, I haven't seen an easy way to provide both options at the same time. Instead of adding that manually, I tend to leave the configuration for TLS and rather document how to set up connections from the client with the truststore. A third option would be to create two different deployment profiles, each deploying different configurations of Kafka: one with TLS, the other without. WDYT?

b-abel:
> I tend to leave the configuration for TLS and rather document how to set up connections from the client with the truststore (I would add a new _Getting Started_ guide for Kafka-based messaging to Hono's documentation).

The PR https://github.com/eclipse/hono/pull/2682 adds a _Getting Started_ guide for Kafka-based messaging to Hono. It shows the required configuration of the Hono CLI to connect to the Kafka broker if it is configured for TLS only.

calohmn:
When using minikube, `minikube tunnel` has to be running when installing the chart.

sophokles73:
> When using minikube, minikube tunnel has to be running when installing the chart.

The chart itself can also use Node Ports instead of using LoadBalancer. We explicitly added the *useLoadBalancer* property for that purpose. I therefore would prefer, if we could also use kafka with Node Ports only, if possible.

b-abel:
> We explicitly added the _useLoadBalancer_ property for that purpose. I therefore would prefer, if we could also use kafka with Node Ports only (based on the value of the property), if possible.

It is quite easy to configure the Bitnami Kafka chart to use Node Ports instead of _LoadBalancer_. What seems to be problematic is changing this depending on the _useLoadBalancer_ property. The reason is that the Bitnami chart is configured in the file `hono/values.yaml`. Helm does not support templating in `value.yaml` files (for details see for example https://github.com/helm/helm/issues/2492 and https://github.com/helm/helm/issues/2133). 
Unless someone comes up with a better workaround, the only way I currently see is to set the property `kafka.externalAccess.service.type` together with `useLoadBalancer`.

charts/hono/example/create_certs.sh
```diff
@@ -41,6 +41,9 @@ AMQP_ADAPTER_KEY_STORE=amqpKeyStore.p12
 AMQP_ADAPTER_KEY_STORE_PWD=amqpkeys
 EXAMPLE_GATEWAY_KEY_STORE=exampleGatewayKeyStore.p12
 EXAMPLE_GATEWAY_KEY_STORE_PWD=examplegatewaykeys
+KAFKA_KEY_STORE=kafkaKeyStore.jks
+# the bitnami Kafka chart expects truststore and keystore to have the same name
+KAFKA_KEY_STORE_PWD=honotrust
```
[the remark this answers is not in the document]
b-abel: Oh yes :scream:

charts/hono/example/ca_opts
```diff
@@ -146,3 +146,10 @@ subjectKeyIdentifier = hash
 keyUsage             = keyAgreement,keyEncipherment,digitalSignature
 extendedKeyUsage     = serverAuth, clientAuth
 subjectAltName       = DNS.1:localhost
+
+[ req_ext_kafka ]
+
+subjectKeyIdentifier = hash
+keyUsage             = keyAgreement,keyEncipherment,digitalSignature
+extendedKeyUsage     = serverAuth, clientAuth
+subjectAltName       = DNS.1:*.eclipse-hono-kafka-headless,DNS.2:*.eclipse-hono-kafka-headless.hono,DNS.3:*.eclipse-hono-kafka-headless.hono.svc.cluster.local,DNS.4:localhost
```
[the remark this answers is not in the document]
b-abel: I think so. The idea here is that clients directly address a pod using the headless service. The address is then `<release-name>-0.<release-name>-headless.<namespace>.svc.cluster.local`. The wildcard matches the pod name (would even work for multiple brokers).
But looking at this, I found another problem: the release names are part of the service names, and they are determined only during deployment. I created an issue for this: #252.

b-abel:
> Unless someone comes up with a better workaround, the only way I currently see is to set the property `kafka.externalAccess.service.type` together with `useLoadBalancer`.

@sophokles73 I added a check that lets the deployment fail if those two properties are not consistent with each other.

charts/hono/README.md
```diff
@@ -529,3 +529,65 @@ helm install --dependency-update -n hono --set honoImagesType=quarkus-native ecl
...
+
+## Using Kafka based Messaging
+
+The chart can be configured to use Kafka as the messaging network instead of an AMQP 1.0 messaging network.
+The property `messagingNetworkType` is used to select the type of the messaging network.
+
+The following command provides a quickstart for Kafka based messaging:
+
+```bash
+helm install --dependency-update -n hono --set messagingNetworkType=kafka --set kafkaMessagingClusterExample.enabled=true --set amqpMessagingNetworkExample.enabled=false  eclipse-hono eclipse-iot/hono
```
calohmn, line 542: How about adding a note here, that `minikube tunnel` needs to be running if using minikube.
I can imagine this being a common mistake when using minikube, with the install command running into a timeout in that case.

b-abel: Done (and rebased).

calohmn (approved):
LGTM

charts/hono/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-secret.yaml
```diff
@@ -41,4 +41,5 @@ data:
   cert.pem: {{ .Files.Get "example/certs/amqp-adapter-cert.pem" | b64enc }}
   trusted-certs.pem: {{ .Files.Get "example/certs/trusted-certs.pem" | b64enc }}
   adapter.credentials: {{ .Files.Get "example/amqp-adapter.credentials" | b64enc }}
-{{- end }}
\ No newline at end of file
+  truststore.jks: {{ .Files.Get "example/certs/trustStore.jks" | b64enc }}
```
sophokles73, line 44: what do we need this for? We already configure the *trusted-certs.pem* which contains all the trusted certificates ....

b-abel: Kafka does not support _pem_ files before version 2.7. Hono uses the Kafka clients in version 2.6.

sophokles73: But Hono also supports JKS files so FMPOV we should either use JKS *or* PEM files, but not both ...

b-abel: I can provide a PR that replaces the `trusted-certs.pem` with a JKS file if you like.

sophokles73: with *a JKS* file? My understanding is that we should be able to use *only* the `truststore.jks` file or does that contain other certs than `trusted-certs.pem`?

b-abel: Right, `trustStore.jks` is the JKS file to be used everywhere instead of `trusted-certs.pem`.

sophokles73: I think you should indeed change it that way.

b-abel: @sophokles73 Please have a look at #257.

sophokles73: > Kafka does not support pem files before version 2.7. Hono uses the Kafka clients in version 2.6.

Is there a particular reason why we do not simply use the 2.7.x client and continue to use our PEM files?

b-abel: > > Kafka does not support pem files before version 2.7. Hono uses the Kafka clients in version 2.6.
> 
> Is there a particular reason why we do not simply use the 2.7.x client and continue to use our PEM files?

Hono does not directly use the Kafka clients, but they are a transitive dependency of the Vert.x Kafka client. Currently, it is version `3.9.7` of `vertx-kafka-client` which depends on version `2.6.0` of the Kafka clients.

Another minor point might be client/broker compatibility. Kafka version 2.7 is pretty recent. While I find it hard to find general reliable information on that topic, it seems advisable to use brokers with an equal or higher version than the clients. So, this would potentially increase the minimal Kafka version required by Hono.

PR #240, opened 2021-05-19
[#239] Document required Helm version to be 3.1 or later

### LINKED ISSUE #239: Define Helm 3.1 as required version
As discussed in the community call:
Newer bitnami charts used in the requirements of the IoT Packages charts require Helm 3.1 (see https://github.com/eclipse/packages/pull/218#issuecomment-820566813). Therefore we should document Helm 3.1 or later as the required version and update the GitHub Action workflow to use a chart-testing-action with support for Helm 3.1.

calohmn:
Also update GitHub Action workflow to support Helm 3.1 charts.
This fixes #239.


sophokles73 (approved):
LGTM

PR #241, opened 2021-05-19
Remove support for outdated kubernetes versions

sophokles73:
Drops support for k8s < 1.7.0.
Also updates the c2e package to use latest Hono 1.7.x version and use Command Router by default


sophokles73:
closing in favor of #242

PR #242, opened 2021-05-20
Fix CI build

sophokles73:
The chart-testing-action version 2 has introduced some breaking changes.
in particular, it now installs the chart testing tool instead of
wrapping it. Thus invocation of the `ct` command line tool needs to be
adapted.

Also changed the `kubeval.sh` script to no longer install helm as this is now
done by the `ci.yml` workflow.

Also removed outdated k8s versions from the install-chart job.

### DIFF
--- a/.github/kubeval.sh
+++ b/.github/kubeval.sh
@@ -3,7 +3,7 @@
 # use kubeval to validate helm generated kubernetes manifest
 #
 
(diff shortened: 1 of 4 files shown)


sophokles73:

ctron (approved):
Look good to me. Thanks for the fix!

PR #235, opened 2021-05-14
[#234] Add embedded JDBC-based Device Registry

b-abel:
This resolves #234. It should not be merged before Hono 1.8.0 is released because it depends on [a feature](https://github.com/eclipse/hono/issues/2587) that will be available from this version on.

This PR adds `embedded` as a new type of example device registry. The embedded device registry uses the existing JDBC device registry implementation and activates the application profile `create-schema` on it. The device registration database, as well as the tenant database, are H2 databases stored on a persisted volume.

PR #250, opened 2021-06-01
Request more memory for embedded device registry type

sophokles73:
This improves #235

The JDBC based registry, when used with an embedded H2 DB, will require
more memory to accommodate the additional H2 DB functionality.

The resource requests and limits of the device registry can now also be
set at the particular registry type level, falling back to the default
values specified at the registry example level.

### DIFF
--- a/charts/hono/values.yaml
+++ b/charts/hono/values.yaml
@@ -1228,6 +1228,10 @@ deviceRegistryExample:
     # configuration property of the component's readiness probe.
     # The value of the top level "readinessProbeInitialDelaySeconds" property will be used if not set.
     readinessProbeInitialDelaySeconds:
+    # If not specified here, then the values from "deviceRegistryExample.resources" are used.
+    # Refer to https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
+    # for a description of the properties' semantics.
+    resources:
 
   # embeddedJdbcDeviceRegistry contains configuration properties specific to the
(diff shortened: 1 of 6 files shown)


charts/hono/values.yaml
```diff
@@ -1257,6 +1261,18 @@ deviceRegistryExample:
...
     readinessProbeInitialDelaySeconds:
+    # resources contains the container's requests and limits for CPU and memory
+    # as defined by the Kubernetes API.
+    # If not specified here, then the values from "deviceRegistryExample.resources" are used.
+    # Refer to https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
+    # for a description of the properties' semantics.
+    resources:
+      requests:
+        cpu: "200m"
+        memory: "400Mi"
```
calohmn, line 1272: Isn't 300Mi (as for the protocol adapters) enough here? `kubectl top pod` showed 263Mi for the pod initially when I tested this.
Just saying, because with 400Mi here, we have 2072Mi for the complete standard deployment already.

sophokles73: With 300Mi the integration tests of Hono when run with the embedded H2 DB sometimes fail, i.e. the registry container crashes ...

calohmn: Oh, ok.

calohmn (approved):
LGTM

PR #256, opened 2021-06-07
Extend Ditto logging configuration options

stmaute:
With this PR it is possible to configure Ditto to log to a logstash endpoint.
It is also possible to pipe logs into log files.
Add initContainers to Ditto deployments in order to write log files in the _/var/log/dittto_ directory on the host.
Add initContainer to nginx deployment to wait until the gateway service is ready.


thjaeckle (approved):
LGTM 👍

PR #257, opened 2021-06-07
[Hono] Replace usage of pem truststore by JKS

b-abel:
As discussed in https://github.com/eclipse/packages/pull/237#discussion_r646366566:
Kafka (in the version currently used by Hono) does not support PEM files. Instead of providing the same certificates in PEM as well as in JKS format to Hono's components, this replaces the usage of the file `trusted-certs.pem` with the file `trustStore.jks`. The JKS file requires a password which is added to the configuration as well.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.8.4
+version: 1.8.5
 # Version of Hono being deployed by the chart
 appVersion: 1.8.0
 keywords:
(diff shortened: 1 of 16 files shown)


b-abel:
What I don't like about this PR: the JKS file requires a password which is now hardcoded in many places in the chart. While it would be possible to reduce the number of occurrences by referencing a property, we will hardly achieve having only a single place where it can be changed.
Are we sure that this change is worth the effort?

sophokles73:
> What I don't like about this PR: the JKS file requires a password

Didn't the JKS file already require one? The JKS file has not been introduced by this PR, has it? It had been introduced in #237, hadn't it?

b-abel:
> > What I don't like about this PR: the JKS file requires a password
> 
> Didn't the JKS file already require one? The JKS file has not been introduced by this PR, has it? It had been introduced in #237, hadn't it?

Sorry, that was misleading. The JKS file has always needed one (even before #237). FMPOV this is a good thing. But the JKS file is now used referenced in the `application.yaml` files of Hono's components and there the password needs to be provided every time as well. That was not the case with the PEM file because it does not need a password. So, the security is not worse than before, from a usability perspective it is just a stumbling block for users who change the password of the JKS when creating new certificates.

b-abel:
I would base a new solution on https://github.com/eclipse/hono/issues/2731.

PR #259, opened 2021-06-14
[Hono] Add NodePort deployment to Hono readme

b-abel:
This adds documentation of the properties that are required to use services of type 'NodePort' instead of type 'LoadBalancer' to the readme of the Hono chart.

### DIFF
--- a/charts/hono/README.md
+++ b/charts/hono/README.md
@@ -546,8 +546,11 @@ The following command provides a quickstart for Kafka based messaging (ensure `m
 helm install --dependency-update -n hono --set messagingNetworkTypes[0]=kafka --set kafkaMessagingClusterExample.enabled=true --set amqpMessagingNetworkExample.enabled=false eclipse-hono eclipse-iot/hono
 ```
 
-It enables the deployment of an example Kafka cluster, disables the deployment of the AMQP 1.0 messaging network 
-and configures adapters and services to use Kafka based messaging.
+The parameters enable the deployment of an example Kafka cluster, disable the deployment of the AMQP 1.0 messaging network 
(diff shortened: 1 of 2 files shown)


charts/hono/README.md
```diff
@@ -536,14 +536,17 @@ of Quarkus based services images.
 The chart can be configured to use Kafka as the messaging network instead of an AMQP 1.0 messaging network.
 The property `messagingNetworkType` is used to select the type of the messaging network.
 
-The following command provides a quickstart for Kafka based messaging (ensure `minikube tunnel` is running when using Minikube):
+The following command provides a quickstart for Kafka based messaging when using Minikube (ensure `minikube tunnel` is running ):
```
calohmn, line 539: Is this change needed? There is no minikube requirement here, or is there?

b-abel: This does not only work on Minikube but it won't work for example on microk8s out of the box. _If_ you use Minikube, as it assumed for most of the examples, then you can use this command, if not, you might have to use another configuration. Before it sounded like you only have special requirements on Minikube but on other Kubernetes distributions it would just work without additional steps.

calohmn: It's just that I think that people wanting to deploy Hono on some cloud-based Kubernetes cluster will not feel addressed here when explicitly only mentioning minikube. Therefore I would rather keep this generic and maybe add a note regarding microk8s or other distributions on what to take care of regarding the configuration.

b-abel: I'll just revert this line. The hint about using it without a load balancer comes two sentences later anyway.

b-abel:
Rebased.

calohmn (approved):
LGTM

PR #262, opened 2021-06-16
Hono: Allow to disable TLS on example Kafka deployment

b-abel:
Adds client configuration for cases in which the example Kafka broker is configured to allow unencrypted connections from clients. This is intended to be used in the Hono Sandbox.


calohmn (approved):
LGTM

PR #263, opened 2021-06-16
Hono: Enable parallel usage of AMQP and Kafka messaging

b-abel:
Instead of selecting one messaging type with the property `messagingNetworkType`, the messaging configuration is now selected with the property `messagingNetworkTypes` that allows setting more than one value. This is intended to be used for the deployment of Hono's sandbox, where both messaging networks should be enabled simultaneously.

### DIFF
--- a/charts/hono/README.md
+++ b/charts/hono/README.md
@@ -534,12 +534,16 @@ of Quarkus based services images.
 ## Using Kafka based Messaging
 
 The chart can be configured to use Kafka as the messaging network instead of an AMQP 1.0 messaging network.
-The property `messagingNetworkType` is used to select the type of the messaging network.
+The configuration `messagingNetworkTypes[0]=kafka` deploys Hono configured to use Kafka for messaging.
+It is possible to enable AMQP _and_ Kafka based messaging at the same time (command line parameters of the deployment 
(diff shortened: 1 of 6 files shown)


charts/hono/README.md
```diff
@@ -534,20 +534,24 @@ of Quarkus based services images.
 ## Using Kafka based Messaging
 
 The chart can be configured to use Kafka as the messaging network instead of an AMQP 1.0 messaging network.
-The property `messagingNetworkType` is used to select the type of the messaging network.
+The configuration `messagingNetworkTypes[0]=kafka` deploys Hono configured to use Kafka for messaging.
+It is possible to enable AMQP _and_ Kafka based messaging at the same time (command line parameters of the deployment 
+would be `--set messagingNetworkTypes[0]=amqp --set messagingNetworkTypes[1]=kafka`). Each tenant in Hono can then be
```
calohmn, line 539: How about using `--set "messagingNetworkTypes={amqp,kafka}"` instead?

b-abel: This syntax is more error-prone FMPOV. If you forget to quote the property, it will fail on the command line. All other command line parameters in this document do not need quotation.

calohmn: yes, ok.

calohmn (approved):

PR #264, opened 2021-06-24
[Hono] Do not expose Command Router and Device Connection Service.

b-abel:
The "Command Router" and "Device Connection Service" components are no longer exposed to the outside of the Kubernetes cluster.

PS: I just saw that I pushed by accident a very small fix for the Auth service config. I'll leave it here because it's really small and obvious :)


b-abel:
Btw: Exposing the Command Router has an ugly side-effect: with some LoadBalancer implementations, like the one that k3s uses, the exposed services share the same IP address. This leads to a port conflict between the Command Router and the AMQP Protocol Adapter.

calohmn (approved):
LGTM

PR #265, opened 2021-06-29
[Hono] Add the Let's Encrypt root cert to trusted-certs.pem

b-abel:
Adds the root certificate (https://letsencrypt.org/certs/isrgrootx1.pem) of Let's Encrypt to the example trust store. This allows deploying setups where the example certificates are used together with certificates issued by the Let's Encrypt CA.


charts/hono/example/create_certs.sh
```diff
@@ -105,7 +105,7 @@ openssl req -config ca_opts -reqexts intermediate_ext -new -key $DIR/ca-key.pem
 
 echo ""
 echo "creating PEM trust store ($DIR/trusted-certs.pem) containing CA certificate"
-cat $DIR/ca-cert.pem $DIR/root-cert.pem > $DIR/trusted-certs.pem
+cat $DIR/ca-cert.pem $DIR/root-cert.pem lets-encrypt-root-cert.pem > $DIR/trusted-certs.pem
```
sophokles73, line 108: maybe we can download the cert from the Let's Encrypt site instead? That would relief us from having to maintain a copy in the repo ...

calohmn (commented):
Shouldn't also the intermediate certificate be added to `trusted-certs.pem`, i.e. R3 from [here](https://letsencrypt.org/certificates/)?

b-abel:
The script downloads now the LE root and intermediate certificate and adds them to the trust store.

sophokles73 (approved):
LGTM

sophokles73:
@calohmn WDYT?

calohmn (approved):
LGTM

PR #268, opened 2021-07-15
[#267] Hono: Use standard Kafka port for external access.

### LINKED ISSUE #267: [Hono] Use the standard Kafka port for external access
Currently, when deploying the Hono chart with the example Kafka cluster, the standard Kafka port 9092 is used internally inside the Kubernetes cluster and for external access the port 9094 is exposed. This is the default behavior of the Bitnami Kafka chart that Hono uses. From the POV of a typical Hono user, it would be better to expose the standard port. 
I propose to change the configuration so that the internal and the external ports are simply switched.

b-abel:
This fixes #267. 
It configures the example Kafka cluster that can be deployed with Hono to expose the standard Kafka port 9092 for external access. Before the port 9094 was used as an external port and 9092 was used only internally in the Kubernetes cluster. This commit changes both ports.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.9.5
+version: 1.9.6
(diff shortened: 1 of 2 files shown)


b-abel:
@sophokles73, does that fit with what you're expecting?

sophokles73 (approved):
LGTM

PR #275, opened 2021-08-04
[#249] Hono: Make Device registry with "embedded" DB the default

### LINKED ISSUE #249: [Hono] Make Device Registry with embedded DB the Default
Hono 1.9.0 deprecates the file-based device registry implementation which is currently used as the default of the Hono chart. This could be replaced by the JDBC-based implementation with an embedded database, which is already supported with #235. 
Technically, this only requires changing the property "deviceRegistryExample.type" to "embedded" in the file `values.yaml`. 

The readme file should be updated as well. For instance, it states that the file-based implementation is used as the default. The part about the steps required to use a production-grade device registry should be updated. The part about the _Device Connection API_ should be checked as well if it contains assumptions about the default device registry which are no longer valid.

b-abel:
This fixes #249.

### DIFF
--- a/charts/hono/README.md
+++ b/charts/hono/README.md
@@ -409,13 +409,13 @@ The Device Connection API is deprecated and will be replaced by the Command Rout
 In order to use the Device Connection API, the *useCommandRouter* property has to be set to `false` when deploying the Helm chart.
 
 ```bash
-helm install --dependency-update -n hono --set useCommandRouter=false eclipse-hono eclipse-iot/hono 
+helm install --dependency-update -n hono --set useCommandRouter=false --set deviceRegistryExample.type=file eclipse-hono eclipse-iot/hono 
 ```
 
 #### Example Implementation
 
-Hono's example Device Registry component contains a simple in-memory implementation of the Device Connection API.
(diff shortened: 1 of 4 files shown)


charts/hono/README.md
```diff
@@ -414,8 +414,8 @@ helm install --dependency-update -n hono --set useCommandRouter=false eclipse-ho
 
 #### Example Implementation
```
calohmn, line 415: The above command `helm install --dependency-update -n hono --set useCommandRouter=false eclipse-hono eclipse-iot/hono` needs to be adapted so that `--set deviceRegistryExample.type=file` is in there.
Otherwise the command will fail:
```
Error: execution error at (hono/templates/hono-adapter-mqtt/hono-adapter-mqtt-vertx-secret.yaml:41:10): .Values.deviceRegistryExample.type MUST be 'file' and .Values.deviceConnectionService.enabled set to true if example Device Registry is disabled and no other Device Connection service is configured
```

b-abel: Right. That is basically what the following paragraph explains.

calohmn (approved):
LGTM

PR #278, opened 2021-08-05
Hono: Set node port for external Kafka service.

b-abel:
This defines a node port for the external service of the example Kafka broker.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.9.11
+version: 1.9.12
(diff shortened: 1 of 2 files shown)


calohmn:
@b-abel Can you rebase and update the chart version?

calohmn:
@b-abel What is the motivation here?

b-abel:
The idea is to define a fixed port number for the external Kafka service. Without this setting, Kubernetes picks port and there is no guarantee that it will always be the same number. This is only relevant if the service is of type "NodePort" (and not "LoadBalancer"). The other services in Hono define the ports as well.

calohmn:
> The other services in Hono define the ports as well.

ok, right.

>  This is only relevant if the service is of type "NodePort" (and not "LoadBalancer"). 

It is also relevant for type "LoadBalancer", the LoadBalancer Type being an extension of the the NodePort type. "LoadBalancer" is used in the config changed here, anyway.

calohmn (approved):
LGTM

PR #286, opened 2021-09-06
Fix rate()/irate() Hono dashboard queries

calohmn:
- Replace `$__range` with `$__rate_interval`. 
  If a Hono protocol adapter pod gets deleted and a new one gets started, a large chart time range would otherwise result in an  equally large period being displayed with wrong (doubled) chart values.
- Replace `$__interval` with `$__rate_interval`. 
 This prevents empty rate/irate() results in case of a short range (with not enough data points in `$__interval`) being displayed.


calohmn:
@sophokles73 Could you have a look?

sophokles73 (approved):
LGTM

Issue #274, opened 2021-07-22
New package proposal


ctron:
I would like to propose a new package, consisting of the following stack:

* Device/Sensor
  * Drogue IoT Device
* Gateway
  * Eclipse ioFog (as the edge platform)
  * Eclipse Kura (as the IoT workload on the edge)
* Cloud
  * Drogue IoT Cloud (as the IoT connectivity layer)
  * Apache Kafka (as the event streaming solution)
  * Knative Eventing (as the event orchestration)
  * Eclipse Ditto (as digital twin component)
  * Eclipse Streamsheets (as data visualization solution)

The goal of the package is to show an end-to-end (device-to-cloud) architecture with open source components, focusing on the use-case of telemetry data acquisition.

We are willing to drive this project, but would hope that the respective projects would be supportive in this effort.

We also did some initial tests, to check out if this is technically feasible. And so far, this looks good. There are a few tweaks we had to do on the different components, but we hope that these tweaks will land in the respective projects, rather than the IoT Packages project.

Regarding the device and edge layer, we plan two approaches:

1.) Use a real microcontroller (STM32 or nRF) and and a Raspberry Pi (Zero W or better)
2.) Emulate the micrcontroller using WebAssembly in the browser and the edge node using a virtual machine

As a first step, I am asking for feedback on this proposal. If you like please indicate that you are interested in supporting this effort.

thjaeckle:
Hi @ctron 
Is this something still intended to be executed on a local machine? Seems for me like a "cloud only" setup.

I could support in the Ditto part of such a package and would suggest to push data into Streamsheets utilizing the "Digital Twin" of Ditto (e.g. via HTTP push, doing conversion to a cloud events inside of a custom Ditto payload mapper) in order to work on a normalized data structure (the Ditto "Thing" and Ditto Protocol events).

ctron:
> Is this something still intended to be executed on a local machine? Seems for me like a "cloud only" setup.

Yes, it should be possible to test this on your local machine. It still requires "Kubernetes" and "Helm" as intended for the IoT Packages.

You might need a bit of RAM to run this. But aside from Kafka, Keycloak and Ditto, the different components don't need much. Not adding Ditto, we can fire this up on a GitHub actions runner, which has 7.5GiB RAM available.

What might be a bit tricky is the "non-cloud" part of the edge node and the microcontroller for the sensor. Thus the idea to emulate this.

[...]

PR #288, opened 2021-09-17
Add "Telemetry E2E" package

ctron:
This is a draft PR for the new "Telemetry E2" package. As outlined in #274

The current state is:

* [ ] Drogue IoT Device – Pending
* [ ] Eclipse ioFog – Issues with Kubernetes deployment: waiting for next 3.x pre-release
* [ ] Eclipse Kura – Performance issue in armv7 container: investigating Java 11 from AdoptOpenJDK
* [x] Drogue IoT Cloud
* [x] Eclipse Ditto
* [x] Eclipse Streamsheets

Additional tasks:

* [ ] Documentation

Changes from original plan:

* Drop the idea of using Raspberry Pi Zero: Support for armv7 is already challenging, the Zero however is an armv6 platform, which is even less supported in the Java world. Thus is may be too much effort this get this working properly, and switching to a Raspberry Pi 3+ model seems a reasonable compromise.

My proposal is to continue working on the documentation of the cloud-side parts, add a Kura emulator to the Kubernetes deployment for now and do a first "preview" (draft) release once this is accomplished.


ctron:
@calohmn I think it would be great if you (as project lead) could give a quick thumbs up (or comment otherwise) on the overall idea of adding this chart.

packages/telemetry-e2e/values.yaml
```diff
@@ -0,0 +1,182 @@
...
+        requests:
+          memory: 128Mi
+          cpu: "100m"
+        limits:
+          memory: 256Mi
+          cpu: "250m"
+
+ditto:
+  fullnameOverride: ditto
+  dittoTag: 2.1.0-M3
```
thjaeckle, line 105: this can be dropped now in favor of using the latest 2.1 Ditto chart with final Ditto 2.1.0 in it

ctron:
I squashed the commits and extracted the homepage parts to another PR.

Unless there are any concerns, I would release a first alpha version of this chart in the next days.

PR #294, opened 2021-10-04
update tool versions

ctron:
Noticing a "segmentation fault" on the CI run of PR #288, I checked our current tool versions and found e.g. Helm at 3.1 (current version 3.7).

I went ahead and updated the Helm versions in our GitHub workflow: a) in the hope this fixes the segmentation fault b) because I think it is time to (conservatively) update to a newer version.


thjaeckle (approved):
I also recently thought that we are quite outdated here, so I appreciate the update 👍

PR #296, opened 2021-10-04
#295: make ditto-nginx init container image configurable

### LINKED ISSUE #295: make ditto-nginx init container image configurable
Currently, the image of the nginx init container is hardcoded: 

https://github.com/eclipse/packages/blob/d94c358e02d82d42357865606395a9407ef7353b/charts/ditto/templates/nginx-deployment.yaml#L51

Therefore, make this configurable in order to allow pulling the image from another registry.

JeffreyThijs:
Closes #295.


thjaeckle (approved):
LGTM, thx for the contribution 👍

PR #297, opened 2021-10-05
fixed Ditto systemProps was omitting double quotes

thjaeckle:
* switched from "toYaml" to "range"

Probably caused https://github.com/eclipse/ditto/issues/1201


dguggemos (approved):
LGTM 👍

PR #298, opened 2021-10-05
ability to configure truststore for (example device registry and comm…

BobClaerhout:
…and router) clients in hono


charts/hono/values.yaml
```diff
@@ -1217,6 +1217,9 @@ deviceRegistryExample:
     #    secretName: "example-data"
     #    mountPath: "/etc/data"
 
+  # The path to the truststore for clients using this device registry.
```
sophokles73, line 1220: Can you please add some wording that explains under which circumstances this value will be used and for what?

charts/hono/values.yaml
```diff
@@ -1701,6 +1704,9 @@ commandRouterService:
   # The configMap to get additional environment variables from for this command router service.
   envConfigMap:
 
+  # The path to the truststore for clients using this command router service.
```
sophokles73, line 1707: Can you please add some wording that explains under which circumstances this value will be used and for what?

PR #299, opened 2021-10-06
Update k8s versions up to 1.22, keeping the last 5 active

ctron:
First of all I was adding Kubernetes 1.22.x, keeping the 5 latest versions.

Second I noticed that we've been missing validation starting with version 1.19. It looks like the original repository with the schemas doesn't get updated anymore. But there is a new one, that is.


sophokles73 (approved):
LGTM

PR #301, opened 2021-10-07
allow configuring service type

BobClaerhout:


charts/hono/templates/_helpers.tpl
```diff
@@ -637,6 +637,9 @@ The scope passed in is expected to be a dict with keys
 Adds port type declarations to a component's service spec.
 */}}
 {{- define "hono.serviceType" }}
+{{- if .Values.serviceType }}
+  type: {{ .Values.serviceType }}
+{{- else }}
```
sophokles73, line 642: I guess we can turn this into an `elseif` as well, right?

charts/hono/templates/jaeger/jaeger-query-service.yaml
```diff
@@ -28,11 +28,15 @@ spec:
       targetPort: query-http
   selector:
     {{- include "hono.matchLabels" $args | nindent 4 }}
+{{- if .Values.serviceType }}
+  type: {{ .Values.serviceType }}
+{{- else }}
```
sophokles73, line 33: elseif?

charts/hono/Chart.yaml
```diff
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.10.4
+version: 1.10.5
```
sophokles73, line 18: this needs to be bumped to 1.10.6

charts/hono/templates/_helpers.tpl
BobClaerhout: updated

PR #304, opened 2021-10-11
Add RBAC role for Hono command router to list/watch pods

calohmn:
This adds the RBAC role needed for the Hono Command Router `AdapterInstancesLivenessService` introduced in Hono 1.10.0 (see https://github.com/eclipse/hono/issues/2028).

### DIFF
--- a/charts/hono/templates/hono-service-command-router/hono-service-command-router-role.yaml
+++ b/charts/hono/templates/hono-service-command-router/hono-service-command-router-role.yaml
@@ -1,0 +1,11 @@
+{{- if and .Values.useCommandRouter .Values.commandRouterService.enabled }}
+{{- $args := dict "dot" . "name" "service-command-router" }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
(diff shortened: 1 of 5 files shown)


charts/hono/templates/hono-listPods-role.yaml
```diff
@@ -0,0 +1,9 @@
+{{- $args := dict "dot" . "name" "listPods" }}
```
sophokles73, line 1: is there a particular reason why this role definition is not in the command router folder as well?

charts/hono/templates/hono-service-command-router/hono-service-command-router-role.yaml
```diff
@@ -0,0 +1,11 @@
...
+{{- $args := dict "dot" . "name" "service-command-router" }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  {{- include "hono.metadata" $args | nindent 2 }}
+rules:
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["list", "watch"]
+{{- end }}
```
sophokles73, line 11: missing EOL

PR #306, opened 2021-10-12
Fix for regression in Jenkins Kubernetes plugin

fredg02:
Signed-off-by: Frederic Gurr <frederic.gurr@eclipse-foundation.org>


sophokles73:
@fredg02 is this a fix we need to apply in general to all our build definition files that use custom images for the build?

PR #315, opened 2021-11-09
Fix link to CNCF's kubernetes platform options page

sophokles73:
Also added line breaks.

### DIFF
--- a/homepage/prereqs.md
+++ b/homepage/prereqs.md
@@ -331,7 +331,7 @@ The required Helm version is 3.4 or later.
 
 ### Repository
 
-The Eclipse IoT Packages projects hosts a Helm chart repository for Eclipse IoT projects.
+The Eclipse IoT Packages project hosts a Helm chart repository for Eclipse IoT projects.
 Adding the repository can be done on your local machine be executing:
 
 {% clipboard %}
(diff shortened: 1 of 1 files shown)


homepage/prereqs.md
```diff
@@ -297,8 +331,7 @@ The required Helm version is 3.4 or later.
 
 ### Repository
 
-The Eclipse IoT Packages projects publishes a Helm chart repository for Eclipse IoT projects.
-
+The Eclipse IoT Packages projects hosts a Helm chart repository for Eclipse IoT projects.
```
calohmn, line 334: ` IoT Packages project` - singular

ctron:
I am struggling a bit with the microk8s section. I think it needs a major overhaul.

It was written only for the Cloud2Edge package, but I think the location it is currently in, is a more generic section.

I think there are two ways to deal with this:

1. Move this to the cloud2edge package entirely
2. Clean up the cloud2edge specific sections, and make it more generic

I would hope the latter.

sophokles73:
@ctron I agree with your assessment. However, this PR was only about fixing the link (and adding some line breaks).
Moving/refactoring the microk8s section should be done in a separate PR FMPOV.

ctron (approved):

PR #317, opened 2021-11-09
Do not install Prometheus and Grafana by default.

### LINKED ISSUE #316: Do not install Prometheus and Grafana by default
The Hono chart currently installs Prometheus and Grafana by default. However, my feeling is that most people do not really take advantage of it. Thus, I propose to not install it anymore by default. This would also reduce the minimum HW requirements for installing the (default) configuration ...

sophokles73:
The README has been updated accordingly.

Fixes #316

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -128,6 +128,10 @@ jobs:
           config: .github/kind-config.yaml
           node_image: kindest/node:${{ matrix.k8s }}
         if: ${{ steps.list-changed.outputs.changed == 'true' }}
+      - name: Describe nodes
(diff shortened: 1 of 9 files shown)


sophokles73:
@calohmn I have fixed the failing deployment tests by means of installing a single protocol adapter per deployment variant only. The underlying kind cluster does not have much CPU available, the k8s components already request 650 milli-CPUs and with @ctron's recent addition of the nginx ingress controller, there simply wasn't enough CPU left anymore to install three adapters together.
Would you mind taking a look now?

calohmn (approved):
LGTM

PR #319, opened 2021-11-17
[Hono] Remove obsolete configuration property

sophokles73:
The "defaultJavaOptions" property had never been used in any of the
templates. Instead, the JDK_JAVA_OPTIONS environment variable set on the
containers had been created with the value of component specific
properties.


sophokles73:

calohmn (approved):
LGTM

PR #320, opened 2021-11-17
[Hono] Enable deletion of Kafka topics

calohmn:
The default Kafka broker configuration is adapted here to allow deletion of topics.
This is relevant for the `hono.command_internal.[adapterInstanceId]` topics, created by Hono for the lifetime of a protocol adapter instance after which they shall get deleted.

Relates to https://github.com/eclipse/hono/issues/2901.


sophokles73 (approved):
LGTM

PR #322, opened 2021-11-19
Fix usage of springLoggingConfig

calohmn:
Wrong property name was used in `_helpers.tpl` and in the documentation.


sophokles73 (approved):
LGTM

PR #323, opened 2021-11-23
Update Ditto Helm chart to released Ditto 2.2.0

thjaeckle:
* simplified README.md of the Ditto chart; added namespace, borrowed from Hono readme


stmaute (approved):
LGTM 👍🏻 
@thjaeckle Can the PR be merged?

PR #326, opened 2021-12-14
Update Infinispan image to 11.0.13.Final

calohmn:
Update Infinispan to version with updated log4j (CVE-2021-44228). Using the non-native image since there are no native versions of the current Infinispan releases.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.10.14
+version: 1.10.15
 # Version of Hono being deployed by the chart
 appVersion: 1.10.0
 keywords:
--- a/charts/hono/templates/example-data-grid/statefulset.yaml
+++ b/charts/hono/templates/example-data-grid/statefulset.yaml
@@ -44,8 +44,21 @@ spec:
           containerPort: 11222
           protocol: TCP
         volumeMounts:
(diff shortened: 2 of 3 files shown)


ctron (approved):

calohmn:
The failed test is because of a missing `log4j2.xml` in the config map (needed by the non-native infinispan server image).

I'm currently checking whether changing the image here is needed at all. It isn't clear yet whether the used _native_ server image is affected by the vulnerability.

LOorts-Aloxy:
Infinispan version 11.0.13 is using version 2.15.0 which is still affected by ([CVE-2021-45046](https://nvd.nist.gov/vuln/detail/CVE-2021-45046)). Maybe it's best to only update when infinispan updates to use log4j 2.16.0?

calohmn:
It seems the native image is not affected by the vulnerability. Nonetheless, I think we can do the update here and then switch again to the native image version which is supposed to be available again with the coming 13.0.4 version (using log4j 2.16.0 then). @sophokles73 WDYT?

sophokles73:
> It seems the native image is not affected by the vulnerability

What information/analysis is this assessment based on?

If this assessment is correct, then I tend to agree with @LOorts-Aloxy that we could wait for 13.0.4 to be released and then do the upgrade in one go.

calohmn:
> What information/analysis is this assessment based on?

Discussions in [Infinispan chat](https://infinispan.zulipchat.com/#narrow/stream/118645-infinispan/topic/log4j.20-.20CVE-2021-44228) and https://github.com/infinispan/infinispan-images/issues/67.

calohmn:
No 100% clear statement there though, so merging this to be extra cautious and get the used version here off the radar. We can see this as an intermediate step en route to the 13.0.4 update, to be done once it is available.

PR #327, opened 2021-12-22
Update to infinispan/server-native:13.0.5

calohmn:
Follow-up of #326:
Update Infinispan image to 13.0.5, using again the native image version.

### DIFF
--- a/charts/hono/values.yaml
+++ b/charts/hono/values.yaml
@@ -188,7 +188,7 @@ dataGridExample:
   enabled: false
   # imageName contains the name (including tag)
   # of the container image to use for the example data grid.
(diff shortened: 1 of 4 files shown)


charts/hono/values.yaml
```diff
@@ -188,7 +188,7 @@ dataGridExample:
   enabled: false
   # imageName contains the name (including tag)
   # of the container image to use for the example data grid.
-  imageName: infinispan/server:11.0.13.Final
+  imageName: infinispan/server-native:13.0.5.Final-1
```
sophokles73, line 191: how about simply using `13.0` as the tag?

sophokles73 (approved):
LGTM

PR #331, opened 2022-01-26
Upgrade Kafka chart

sophokles73:
The dependency on the Kafka chart has been changed to version 14.
The default internal and external ports defined in the Kafka chart are
now being used unaltered.


sophokles73:

PR #332, opened 2022-01-27
Upgrade Mongo DB chart

sophokles73:
Use (most recent) chart that installs Mongo DB 4.4 to reflect what we already use in Hono's unit and integration tests


sophokles73:
@ctron would you mind taking a quick look?

PR #334, opened 2022-01-28
Added imagePullSecrets to hono chart for pulling from private registries

kyberpunk:
This pull request allows to pull images in Hono chart from private registry (e.g. using a proxy registry in enterprise infrastructure). Other charts already provide this option. I've used the approach from default helm template. But it is a valid discussion how to unify it across other charts. I intentionally din't use "global" value from Ditto chart since it conflicts with bitnami dependencies which use the same value with different yaml format.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.10.20
+version: 1.10.21
 # Version of Hono being deployed by the chart
 appVersion: 1.10.1
 keywords:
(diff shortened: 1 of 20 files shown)


sophokles73 (commented):
Thanks for the PR, can you also please increase the patch level version of the chart in the Chart.yaml file?

kyberpunk:
Sure, it's done.

charts/hono/Chart.yaml
```diff
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.10.19
+version: 1.10.20
```
sophokles73, line 18: The chart is already at 1.10.20 so you will need to bump it to 1.10.21 ...

sophokles73 (changes requested):

charts/hono/Chart.yaml
kyberpunk: Ok, I've missed it. It's rebased and fixed with version 1.10.21 now.

sophokles73 (approved):
LGTM

sophokles73:
@kyberpunk thanks for contributing :+1:

Issue #325, opened 2021-12-02
Update Hono to version 1.11


b-abel:
This issue is for tracking changes that will be required when updating Hono to version 1.11 (once it is released).

+ Changes regarding configuration property `hono.kafka.defaultClientIdPrefix`: https://github.com/eclipse/packages/issues/324
+ Configure Kafka clients individually: https://github.com/eclipse/hono/issues/2904
+ Define AMQP addresses with `notification/`prefix as multicast addresses in qdrouterd.json: https://github.com/eclipse/hono/pull/2994

Issue #324, opened 2021-11-25
[Hono] Remove configuration property hono.kafka.defaultClientIdPrefix


b-abel:
The discussion in https://github.com/eclipse/hono/issues/2960 concluded that the configuration property `hono.kafka.defaultClientIdPrefix` is to be removed.

calohmn:
I think we can remove it along with the update to Hono 1.11.0.
Removing it now would lead to client IDs with whitespace (e.g. "Hono AMQP Adapter-internal-cmd-d4378820-8425-4034-90d6-fa10a24a014e"), which is not ideal. For Hono 1.11.0 the format of the IDs shall be changed anyway.

PR #335, opened 2022-01-28
Upgrade to Hono 1.11

sophokles73:
This addresses #324 and #325

Upgraded to Hono 1.11.1

### DIFF
--- a/charts/hono/templates/_helpers.tpl
+++ b/charts/hono/templates/_helpers.tpl
@@ -231,26 +231,31 @@ kafka:
 {{- if .dot.Values.kafkaMessagingClusterExample.enabled }}
   commonClientConfig:
     bootstrap.servers: {{ .dot.Release.Name }}-{{ .dot.Values.kafka.nameOverride }}-0.{{ .dot.Release.Name }}-{{ .dot.Values.kafka.nameOverride }}-headless.{{ .dot.Release.Namespace }}:{{ .dot.Values.kafka.service.port }}
-{{- if eq .dot.Values.kafka.auth.clientProtocol "sasl_tls" }}
+  {{- if eq .dot.Values.kafka.auth.clientProtocol "sasl_tls" }}
     security.protocol: "SASL_SSL"
     sasl.mechanism: "SCRAM-SHA-512"
     sasl.jaas.config: "org.apache.kafka.common.security.scram.ScramLoginModule required username=\"{{ first .dot.Values.kafka.auth.sasl.jaas.clientUsers }}\" password=\"{{ first .dot.Values.kafka.auth.sasl.jaas.clientPasswords }}\";"
     ssl.truststore.location: "/etc/hono/truststore.jks"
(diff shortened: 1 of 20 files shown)


charts/hono/templates/_helpers.tpl
```diff
@@ -221,35 +222,35 @@ Add configuration properties for Kafka based messaging to YAML file.
...
 - (mandatory) "component": the name of the component
+- (optional) "kafkaMessagingSpec": the configuration properties to use for the component's
+                                   Kafka producers/consumers/admin clients
 */}}
 {{- define "hono.kafkaMessagingConfig" -}}
-{{- include "hono.kafkaConfigCheck" (dict "dot" .dot) }}
+{{- include "hono.kafkaConfigCheck" . }}
 kafka:
-  defaultClientIdPrefix: {{ .component }}
 {{- if .dot.Values.kafkaMessagingClusterExample.enabled }}
```
calohmn, line 231: In the `kafkaMessagingClusterExample.enabled` case, it would be better to adopt any non-commonClientConfig properties here from the given `.kafkaMessagingSpec`.

charts/hono/templates/_helpers.tpl
```diff
@@ -221,35 +222,35 @@ Add configuration properties for Kafka based messaging to YAML file.
...
-  {{- required ".Values.kafka.auth.clientProtocol has unsuppported value" nil }}
+  {{- required ".Values.kafka.auth.clientProtocol has unsupported value" nil }}
 {{- end }}
-{{- else if not .dot.Values.adapters.kafkaMessagingSpec }}
-  {{- required ".Values.adapters.kafkaMessagingSpec MUST be provided if example Kafka cluster is disabled" nil }}
-{{- else if not (index .dot.Values.adapters.kafkaMessagingSpec.commonClientConfig "bootstrap.servers") }}
-  {{- required ".Values.adapters.kafkaMessagingSpec.commonClientConfig MUST contain 'bootstrap.servers' if example Kafka cluster is disabled" nil }}
-{{- end }}
-{{- if .dot.Values.adapters.kafkaMessagingSpec }}
-  {{- .dot.Values.adapters.kafkaMessagingSpec | toYaml | nindent 2 }}
```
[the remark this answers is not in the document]
sophokles73: I had also thought about that when I made the changes. I have no strong opinion on it so I will change it according to your proposal.

charts/hono/templates/_helpers.tpl
sophokles73: I would rather keep it simple here. If the example Kafka cluster is being used, then we use a fixed configuration. Otherwise, the `kafkaMessagingClusterExample.enabled` property should be set to `false` and a dedicated cluster should be set up outside the scope of this chart.

sophokles73:
@calohmn I have pushed changes

charts/hono/templates/_helpers.tpl
calohmn: ok

calohmn (approved):
LGTM

PR #336, opened 2022-02-04
Several improvements

sophokles73:
- Updated Infinispan configuration schema to version 13
- Use in-memory data store for embedded cache in Command Router instead
of file-store based cache (which does not work with network file system
anyway)
- Added support for explicitly configuring cache usage in Command Router
- Made information output after successful deployment more specific to
the security configuration used with example Kafka cluster

### DIFF
--- a/charts/hono/values.yaml
+++ b/charts/hono/values.yaml
@@ -1781,6 +1781,15 @@ commandRouterService:
       #  insecurePortEnabled: true
       #  insecurePortBindAddress: "0.0.0.0"
 
+      # cache contains properties configuring the embedded or remote cache that
+      # the Command Router uses for storing routing information.
+      # If not set explicitly here, the remote cache configuration from property "dataGridSpec" will be used.
+      # If that property is empty and "dataGridExample.enabled" is "true", the example data grid is used.
+      # If "dataGridExample.enabled" is "false", an embedded cache with a default configuration will be used.
+      # Note that when using the embedded cache, the Command Router pod does NOT support scaling up to a replica
+      # count > 1
+      cache:
+
     # healthCheck contains configuration properties for the service's
     # health check server as defined by
     # https://www.eclipse.org/hono/docs/admin-guide/monitoring-tracing-config/#health-check-server-configuration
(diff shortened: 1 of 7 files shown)


sophokles73:

charts/hono/values.yaml
```diff
@@ -1781,6 +1781,15 @@ commandRouterService:
...
       #  insecurePortBindAddress: "0.0.0.0"
 
+      # cache contains properties configuring the embedded or remote cache that
+      # the Command Router uses for storing routing information.
+      # If not set explicitly here, the remote cache configuration from property "dataGridSpec" will be used.
+      # If that property is empty and "dataGridExample.enabled" is "true", the example data grid is used.
+      # If "dataGridExample.enabled" is "false", an embedded cache with a default configuration will be used.
+      # Note that when using the embedded cache, the Command Router pod does NOT support scaling up to a replica
+      # count > 1
+      cache:
```
calohmn, line 1791: On the one hand, I see that adding this element here makes the configuration more consistent with the structure defined in the Hono Command Router [configuration guide](https://www.eclipse.org/hono/docs/admin-guide/command-router-config/#data-grid-connection-configuration).
On the other hand having another cache configuration element besides "dataGridSpec" makes things more complex here. I think we could consider deprecating the "dataGridSpec" element and removing it at some point after having removed also the device connection service elements.

In terms of added functionality, the `cache` element here offers the extra possibility to define `cache.embedded.configurationFile` (which wasn't possible before), correct?

Besides: in the values.yaml there are still links to the ISPN 11 api docs (https://docs.jboss.org/infinispan/11.0/apidocs/...). Could be updated to ISPN 13 as well.

sophokles73: That is great feedback, @calohmn :+1: 
I agree that we should deprecate the *dataGridSpec* and remove it later on together with the Device Connection service.

> In terms of added functionality, the cache element here offers the extra possibility to define cache.embedded.configurationFile (which wasn't possible before), correct?

Yes, that was on my mind when adding it.

I will also update the links to the Infinispan docs ...

sophokles73:
@calohmn I have pushed changes ...

calohmn (approved):
LGTM

PR #340, opened 2022-02-08
[#339] Use deprecated property to configure TLS keys

### LINKED ISSUE #339: Kafka broker fails to start when configured to use TLS
The Kafka broker fails to start when it is configured to secure its endpoints with TLS.

sophokles73:
The "auth.tls.existingSecrets" property in the Kafka chart does not work
as documented. Falling back to the (deprecated but working)
"auth.tls.existingSecret" property.

Also updating to latest Hono service release.

Fixes #339


calohmn (approved):
LGTM

PR #343, opened 2022-02-10
[#303] Support setting log level in Quarkus based images

### LINKED ISSUE #303: [Hono] How to configure log level for "quarkus" image type?
Hello,

we try to deploy 1.10.0 version of Hono on kubernetes cluster by using helm.
We use below honoImagesType:
```
honoImagesType: "quarkus"
```
and we use default applicationProfiles
```
applicationProfiles: "dev"
```
but I still can't see debug logs from mqtt-adapter.

I know that if I change the image type to spring, the applicationProfile will take effect, and I can also see the debug level logs.
**But how can we configure log level if we choose quarkus image type?**

thanks!

sophokles73:
The logging level(s) used in the Quarkus based components can now be set
using the "quarkusLoggingProfile" configuration property that is defined
for all components for which a Quarkus based variant exists.

Fixes #303

### DIFF
--- a/charts/hono/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-secret.yaml
+++ b/charts/hono/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-secret.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.adapters.amqp.enabled }}
 #
-# Copyright (c) 2019, 2021 Contributors to the Eclipse Foundation
+# Copyright (c) 2019, 2022 Contributors to the Eclipse Foundation
 #
 # See the NOTICE file(s) distributed with this work for additional
 # information regarding copyright ownership.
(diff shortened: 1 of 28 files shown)


charts/hono/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-secret.yaml
```diff
@@ -18,6 +18,12 @@ metadata:
   {{- include "hono.metadata" $args | nindent 2 }}
 type: Opaque
 stringData:
+  logging-quarkus-prod.yml: |
+    {{- .Files.Get "config/logging-quarkus-prod.yaml" | nindent 4 }}
+  logging-quarkus-dev.yml: |
+    {{- .Files.Get "config/logging-quarkus-dev.yaml" | nindent 4 }}
+  logging-quarkus-trace.yml: |
+    {{- .Files.Get "config/logging-quarkus-trace.yaml" | nindent 4 }}
```
calohmn, line 26: This block could be put inside a
```
{{- if ( contains "quarkus" .Values.honoImagesType ) }}
```
block. Same for the other secret-yamls.

sophokles73: That is true, but does it make a real difference? We are talking about a few bytes ...

calohmn: Right, we can keep it.

calohmn (approved):
LGTM

PR #346, opened 2022-02-16
Use Jolokia HTTP API for k8s probes

### LINKED ISSUE #345: Artemis broker log is flooded with (useless) entries
The Artemis broker log is full with entries indicating an unsuccessful connection attempt on its amqps acceptor.
These stem from the k8s liveness and readiness probes which try to open a TCP connection but then do not engage in the AMQP connection establishment protocol.
I even suspect that this, over time, leads to the container file system running out of space and the container being restarted after it goes OOM.

sophokles73:
The k8s readiness and liveness probes have been changed to no longer use
a TCP connection attempt but to instead query the broker state via the
Jolokia HTTP API.

Fixes #345

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 1.12.1
+version: 1.12.2
 # Version of Hono being deployed by the chart
 appVersion: 1.12.0
 keywords:
--- a/charts/hono/config/artemis/broker.xml
+++ b/charts/hono/config/artemis/broker.xml
@@ -26,7 +26,7 @@
 
       <!-- this could be ASYNCIO or NIO
        -->
-      <journal-type>NIO</journal-type>
+      <journal-type>ASYNCIO</journal-type>
 
       <paging-directory>./data/paging</paging-directory>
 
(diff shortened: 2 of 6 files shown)


tobias-zeptio:
We've also had this issue, but didn't make the change since the acceptors we want to monitor are on TCP port. Does your configuration ensure that the AMQP interface is open?

sophokles73:
@tobias-zeptio Not explicitly. The response for the readiness check does contain a boolean property, though, which indicates if the acceptor is running. I am not sure if it is worth validating its value as part of the check because my guess would be that querying the *amqps* acceptor this way would fail if it cannot be started. But I might be mistaken.

sophokles73:
This is what the broker returns if the acceptor is started:
```json
{
  "request": {
    "mbean": "org.apache.activemq.artemis:broker=\"hono-broker\",component=acceptors,name=\"amqps\"",
    "attribute": "Started",
    "type": "read"
  },
  "value": true,
  "timestamp": 1645167154,
  "status": 200
}
```

tobias-zeptio:
Ah, I can see that. That should be a good indicator, unless there was a mistake made in the port mapping. :)

Which AMQ-number are you solving with this? We have several that are chatty, be most connections seem to originate from outside the cluster. For example our AWS load balancer.

tobias-zeptio:
We filter out some logs, since they don't add much value.

```
# Filter out logs using regex, see "interface ActiveMQServerLogger" in activemq-artemis repository for numbers
# 222061 and 222107 are for client disconnect warnings
# 224096 is for CORE protocol connections attempt, that we have closed in acceptors
# 224088 is for unused connections, could be failing clients or port probing. Nothing we can do about it with public ports.
logger.org.apache.activemq.artemis.core.server.filter=not(match("AMQ(222061|222107|224096|224088)"))
```

calohmn (approved):
LGTM

PR #347, opened 2022-02-18
Upgrade to latest latest monitoring chart minor versions

sophokles73:
Always use latest minor versions of Prometheus and Grafana charts.
Also split up ConfigMap containing Grafana dashboard definitions into
separate ConfigMaps, each containing a single dashboard definition as
recommended by the Grafana chart documentation.


sophokles73:

calohmn (approved):
LGTM

PR #350, opened 2022-02-25
Ditto: Replace enableDummyAuth by enablePreAuthentication

mladBlum:
ENABLE_DUMMY_AUTH was replaced by ENABLE_PRE_AUTHENTICATION.
You can see that here https://github.com/eclipse/ditto/blob/e25c34e044b2d4c0428874942fa75e078c74ce82/gateway/service/src/main/resources/gateway.conf

I replaced ENABLE_DUMMY_AUTH with ENABLE_PRE_AUTHENTICATION in the README.md, values.yaml and gateway-deployment.yaml accordingly.
I don't know if backward compatibility is required nether I found an easy way to introduce it because the env-variables would affect each other.

PR #351, opened 2022-02-25
Fix ditto nginx proxy_set_header spelling of X-Forwarded-User

mladBlum:


mladBlum:
chart version will be bumped after merge of https://github.com/eclipse/packages/pull/350

thjaeckle:
While you are at it, updating the chart version, could you please also update the Ditto version from 2.3.0 to 2.3.2?

mladBlum:
> While you are at it, updating the chart version, could you please also update the Ditto version from 2.3.0 to 2.3.2?

done.

thjaeckle (approved):
Good catch - I did not notice that typo yet.
Thx 👍

PR #353, opened 2022-03-08
Adjustment of Ditto and cloud2edge charts for Openshift deployment

stmaute:
Adjusted config of securityContext  for ditto when deployed into openshift cluster.
Added some profiles for Ditto, Hono and cloud2edge. 
Added template for Ditto Kafka connection;
Fixed some linting errors;


ctron:
Quite a big PR. I haven't tested it, but it looks good to me.

I think the security context stuff could be made more generic. It isn't an OpenShift specific thing. Some Kubernetes clusters requires specific user settings, while OpenShift dynamically assigns them (and requires that no override is set). As there already are "profiles", maybe re-use that and provide reasonable defaults for minikube, kind, openshift and let other users provide their own values? But just more improvement, feel free to ignore it.

sophokles73:
@stmaute thanks for the PR. Could we split up the PR into three separate PRs, e.g. one each for adjustments to the Hono, Ditto and c2e package charts resepectively? I assume that the changes made to the Hono and Ditto charts are meaningful on their own as well, aren't they? That way it would be a little easier to review and process the PRs.

stmaute:
@sophokles73 Sure I can split this PR into one for the adaptions of ditto and one for the extensions of the cloud2edge package.
I only added one profile to the Hono chart which is not worth a separate PR.

PR #355, opened 2022-03-11
Extend Cloud2edge package for openshift 

stmaute:
Add profile to deploy cloud2edge into an Openshift cluster.
Add another profile to run Eclipse Hono with Kafka based messaging.
Add kafka based connection for connecting Hono and Ditto.


stmaute:
@ctron I have tested the cloud2edge package in an Openshift cluster and everything works like expected.

packages/cloud2edge/profileKafkaMessaging.yaml
```diff
@@ -0,0 +1,21 @@
+#
+# Copyright (c) 2021 Contributors to the Eclipse Foundation
```
thjaeckle, line 2: The year should be the creation year, so 2022

packages/cloud2edge/profileOpenshift.yaml
```diff
@@ -0,0 +1,34 @@
+#
+# Copyright (c) 2021 Contributors to the Eclipse Foundation
```
thjaeckle, line 2: The year should be the creation year, so 2022

thjaeckle (changes requested):
Just a little change request regarding newly added files - the copyright header must include the creation year.

charts/hono/profileOpenshift.yaml
```diff
@@ -0,0 +1,25 @@
...
+  podSecurityContext:
+    enabled: false
+  containerSecurityContext:
+    enabled: false
+
+  zookeeper:
+    podSecurityContext:
+      enabled: false
+    containerSecurityContext:
+      enabled: false
```
sophokles73, line 25: missing EOL

packages/cloud2edge/profileOpenshift.yaml
```diff
@@ -0,0 +1,34 @@
...
+      containerSecurityContext:
+        enabled: false
+
+ditto:
+  ## OpenShift deployment
+  openshift:
+    enabled: true
+    ## OpenShift Routes
+    routes:
+      enabled: true
```
sophokles73, line 34: missing EOL

sophokles73 (changes requested):

stmaute:
@thjaeckle @sophokles73 Fixed the year and missing EOL in files.

charts/hono/profileOpenshift.yaml
```diff
@@ -0,0 +1,25 @@
...
```
sophokles73, line 25: this *profile* cannot actually be used when using `helm install` without unpacking the chart locally, can it?

stmaute: the profile can be used in combination with a released chart version.
You can use the profile with the -f option when installing the chart.
E.g. helm install -n hono -f profileOpenshift.yaml hono eclipse-iot/hono

sophokles73: Sure, however, I do not believe that you can use the `-f` switch with resources that are not on your local file system, i.e. using the switch would require to first download and unpack the Hono chart, or am I mistaken?

stmaute: Sure you would need the file locally to apply it with the install command.

sophokles73: Ok, so the c2e package chart does not rely on this file to be present, right?

thjaeckle (approved):

charts/hono/profileOpenshift.yaml
stmaute: No it doesn't. For c2e there is a separate profile file. Only added the hono profile file because didn't want to create a separate PR for one file :)

stmaute: @sophokles73 If there are no more open points? I would merge the PR.

PR #359, opened 2022-03-18
Remove egress from policyType for swaggerui-networkpolicy.yaml 

stmaute:
Because swagger ui definition is loaded from Github during start of container.


thjaeckle (approved):
LGTM 👍

PR #362, opened 2022-03-29
[#361] Remove obsolete PVC

### LINKED ISSUE #361: hono command router volume size in template is too small
The template of   hono service command router volume PVC shall not fixed in 1 Mi.
(packages/charts/hono/templates/hono-service-command-router/hono-service-command-router-pvc.)
As not ALL the kinds of the storage support such a small size.

For example, the NetApp FAS storages could create volumes not smaller than 1Gi.

-----------------
spec:
  accessModes:
  - ReadWriteOnce
  {{- with .Values.commandRouterService.storageClass }}
  storageClassName: {{ . | quote }}
  {{- end }}
  resources:
    requests:
      storage: **1Mi**

sophokles73:
The Command Router does not use a file backed embedded cache any more.
Thus, there is no reason to deploy a PVC.

Fixes #361


calohmn (approved):
LGTM

PR #363, opened 2022-04-08
Use HTTP scheme for health checks and probes

sophokles73:
Using the standard HTTP scheme will allow the chart to be used with Hono
2.0.0 container images as well. That is because Hono 2.0.0 uses SmallRye
Health to spin up an http server hosting the health check and readiness
endpoints and the http server's default configuration does not use HTTPS
and uses other configuration properties than Hono 1.x's custom rolled
health check server.

Using HTTP by default shouldn't be a big issue because the health check
server's endpoints are not exposed via k8s services anyway.

Also updated to Hono 1.12.2


calohmn (approved):
LGTM

PR #364, opened 2022-04-08
fixed Ditto chart swagger-ui not loading the Ditto OpenAPI spec as default

thjaeckle:


charts/ditto/templates/nginx-configmap.yaml
```diff
@@ -155,7 +155,7 @@ data:
         # swagger
         # access API doc on: /apidoc/2
         location /apidoc/ {
-          rewrite ^/apidoc/([0-9])$ $scheme://$http_host/apidoc/ redirect;
+          rewrite ^/apidoc/([0-9])$ $scheme://$http_host/apidoc/?url=/apidoc/openapi/ditto-openapi-$1.yaml redirect;
```
stmaute, line 158: The idea was to download the OpenAPI spec in the initContainer of swagger-ui here: https://github.com/eclipse/packages/blob/eea58376e841db58da59c23a9c4dedbaf400f58e/charts/ditto/templates/swaggerui-deployment.yaml#L56
Maybe we should check why this isn't working?

thjaeckle: The file is correctly downloaded - however the path to that file is not configured so that the swagger-ui loads this definition by default.
By default, it loads the "petstore" definition.

That's why I added the reference to the (already present) downloaded OpenAPI definition.

stmaute: Okay I understand. Then everything looks good to me.

stmaute (approved):

Issue #352, opened 2022-02-28
[Ditto] Update Ditto Helm chart to Ditto 2.4.0


thjaeckle:
Ditto 2.4.0 will change the used Java runtime in its container images from:
* OpenJ9 with Java 11
* to: Temurin with Java 17

As a result, probably some more adjustments (e.g. GC) will have to be done than typically in order to update to the latest Ditto version.

PR #365, opened 2022-04-14
[#352] update Ditto chart to Ditto 2.4.0

thjaeckle:
* and adjusted from OpenJ9 JVM args to Hotspot args
* updated to new logging env variable options

Fixes: #352


stmaute (approved):
LGTM 👍🏻

PR #371, opened 2022-04-29
Switch variable outcome order for syslog parameter

tobias-zeptio:
Signed-off-by: Tobias Månsson <tobias@zept.io>


thjaeckle:
@tobias-zeptio thanks for the contribtion.
You also need to update the chart version in Chart.yaml to 2.5.1

PR #381, opened 2022-06-28
Hono: Add default-container annotation

calohmn:
This sets the default container to be used by "kubectl logs" or "kubectl exec". The `kubectl.kubernetes.io/default-container` annotation has been introduced with Kubernetes 1.21.
Also fixes outdated links in values.yaml documentation.


calohmn:
Since is the first chart update since adding the `Release Notes` README section:
Should we also add patch-version information there? FMPOV, I think it's enough to include major and minor updates there, and maybe extra noteworthy patch-version information. Otherwise the list could get quite long quickly.

sophokles73 (approved):
LGTM

PR #382, opened 2022-07-01
Add support for passing command line arguments to container CMDs

sophokles73:
The Hono chart already supports setting JAVA_OPTIONS to be passed to the
Hono containers. These options can be used to e.g. specify the amount of
memory to use. However, the native executable variants of the Hono
containers ignore the JAVA_OPTIONS but instead require passing options
as command line arguments to set memory limits etc.

The values.yaml now contains a cmdLineArgs property for each component
which can be used to specify an array of Strings to be passed as command
line arguments to the container's CMD.


calohmn (approved):
LGTM

PR #385, opened 2022-07-11
feat: make hono probes configurable

JeffreyThijs:
This change allows to easily configure the liveliness and readiness probe of the pods of all hono services.

### DIFF
--- a/charts/hono/README.md
+++ b/charts/hono/README.md
@@ -103,6 +103,12 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ## Release Notes
 
+### 2.0.4
+
+* *livenessProbeInitialDelaySeconds* and *readinessProbeInitialDelaySeconds* have been deprecated due to the introduction of the
+  *probes* parameter which allows you to configure the HTTP GET probes more specifically instead of just the initialDelaySeconds. The
+  *probes* parameter is configurable globally but can be overwritten per component.
+
 ### 2.0.0
 
 * The chart now uses Helm API version 2 and thus requires at least Helm version 3 to install.
--- a/charts/hono/templates/_helpers.tpl
+++ b/charts/hono/templates/_helpers.tpl
@@ -479,18 +479,31 @@ The scope passed in is expected to be a dict with keys
 - (mandatory) "componentConfig": the component's configuration properties from the values.yaml file
 */}}
 {{- define "hono.component.healthChecks" }}
+{{- $global := .dot.Values -}}
+{{- $component := .componentConfig -}}
+{{- $probes := mergeOverwrite ( $global.probes | deepCopy ) ( $component.probes | default dict | deepCopy ) -}}
+{{- $deprecatedLivenessProbeInitialDelaySeconds := default .dot.Values.livenessProbeInitialDelaySeconds .componentConfig.livenessProbeInitialDelaySeconds -}}
+{{- $deprecatedReadinessProbeInitialDelaySeconds := default .dot.Values.readinessProbeInitialDelaySeconds .componentConfig.readinessProbeInitialDelaySeconds -}} 
 livenessProbe:
   httpGet:
-    path: "/liveness"
-    port: "health"
-    scheme: "HTTP"
-  initialDelaySeconds: {{ default .dot.Values.livenessProbeInitialDelaySeconds .componentConfig.livenessProbeInitialDelaySeconds }}
+    path: {{ $probes.livenessProbe.httpGet.path }}
+    port: {{ $probes.livenessProbe.httpGet.port }}
+    scheme: {{ $probes.livenessProbe.httpGet.scheme }}
+  periodSeconds: {{ $probes.livenessProbe.periodSeconds }}
+  failureThreshold: {{ $probes.livenessProbe.failureThreshold }}
+  initialDelaySeconds: {{ default $probes.livenessProbe.initialDelaySeconds $deprecatedLivenessProbeInitialDelaySeconds }}
+  successThreshold: {{ $probes.livenessProbe.successThreshold }}
+  timeoutSeconds: {{ $probes.livenessProbe.timeoutSeconds }}
 readinessProbe:
   httpGet:
-    path: "/readiness"
-    port: "health"
-    scheme: "HTTP"
-  initialDelaySeconds: {{ default .dot.Values.readinessProbeInitialDelaySeconds .componentConfig.readinessProbeInitialDelaySeconds }}
+    path: {{ $probes.readinessProbe.httpGet.path }}
+    port: {{ $probes.readinessProbe.httpGet.port }}
+    scheme: {{ $probes.readinessProbe.httpGet.scheme }}
+  periodSeconds: {{ $probes.readinessProbe.periodSeconds }}
+  failureThreshold: {{ $probes.readinessProbe.failureThreshold }}
+  initialDelaySeconds: {{ default $probes.readinessProbe.initialDelaySeconds $deprecatedReadinessProbeInitialDelaySeconds }}
+  successThreshold: {{ $probes.readinessProbe.successThreshold }}
+  timeoutSeconds: {{ $probes.readinessProbe.timeoutSeconds }}
 {{- end }}
 
 {{/*
--- a/charts/hono/templates/artemis/artemis-deployment.yaml
+++ b/charts/hono/templates/artemis/artemis-deployment.yaml
@@ -65,21 +65,21 @@ spec:
           {{- . | toYaml | nindent 10 }}
         {{- end }}
         livenessProbe:
-          initialDelaySeconds: {{ $args.componentConfig.livenessProbeInitialDelaySeconds }}
-          periodSeconds: 17
+          initialDelaySeconds: {{ $args.componentConfig.probes.livenessProbe.initialDelaySeconds }}
+          periodSeconds: {{ $args.componentConfig.probes.livenessProbe.periodSeconds }}
+          timeoutSeconds: {{ $args.componentConfig.probes.livenessProbe.timeoutSeconds }}
           exec:
             command:
             - "/bin/bash"
             - {{ printf "%s/liveness-probe.sh" $args.configMountPath | quote }}
-          timeoutSeconds: 3
         readinessProbe:
-          initialDelaySeconds: {{ $args.componentConfig.readinessProbeInitialDelaySeconds }}
-          periodSeconds: 10
+          initialDelaySeconds: {{ $args.componentConfig.probes.readinessProbe.initialDelaySeconds }}
+          periodSeconds: {{ $args.componentConfig.probes.readinessProbe.periodSeconds }}
+          timeoutSeconds: {{ $args.componentConfig.probes.readinessProbe.timeoutSeconds }}
           exec:
             command:
             - "/bin/bash"
             - {{ printf "%s/readiness-probe.sh" $args.configMountPath | quote }}
-          timeoutSeconds: 3
         securityContext:
           privileged: false
         volumeMounts:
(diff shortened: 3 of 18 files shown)


charts/hono/templates/_helpers.tpl
```diff
@@ -479,18 +479,30 @@ The scope passed in is expected to be a dict with keys
...
-    scheme: "HTTP"
-  initialDelaySeconds: {{ default .dot.Values.livenessProbeInitialDelaySeconds .componentConfig.livenessProbeInitialDelaySeconds }}
+    path: {{ $probes.livenessProbe.httpGet.path | default "/liveness" }}
+    port: {{ $probes.livenessProbe.httpGet.port | default "health" }}
+    scheme: {{ $probes.livenessProbe.httpGet.scheme | default "HTTP" }}
+  periodSeconds: {{ $probes.livenessProbe.periodSeconds | default 10 }}
+  failureThreshold: {{ $probes.livenessProbe.failureThreshold | default 3 }}
+  initialDelaySeconds: {{ $probes.livenessProbe.initialDelaySeconds | default 300 }}
+  successThreshold: {{ $probes.livenessProbe.successThreshold | default 1 }}
+  timeoutSeconds: {{ $probes.livenessProbe.timeoutSeconds | default 1 }}
```
sophokles73, line 495: IMHO we should increase the timeout here to e.g. 3 seconds to prevent unnecessary restarts of components, as 1s has proven to be hit quite often with Hono 2.0.0 deployments

sophokles73 (commented):
I like the overall approach :-)
However, in a service release, we cannot simply remove the existing properties for configuring the probe delays. Instead, we should keep them an mark them as deprecated, providing a hint what to use instead. We can, however, keep them empty and implement logic, that uses the old probes if the values are set explicitly, but otherwise uses your new probes based on the values of the new properties in values.yaml.
WDYT?

charts/hono/templates/_helpers.tpl
```diff
@@ -479,18 +479,30 @@ The scope passed in is expected to be a dict with keys
...
+{{- $component := .componentConfig -}}
+{{- $probes := mergeOverwrite ( $global.probes | deepCopy ) ( $component.probes | default dict | deepCopy ) -}}
+
 livenessProbe:
   httpGet:
-    path: "/liveness"
-    port: "health"
-    scheme: "HTTP"
-  initialDelaySeconds: {{ default .dot.Values.livenessProbeInitialDelaySeconds .componentConfig.livenessProbeInitialDelaySeconds }}
+    path: {{ $probes.livenessProbe.httpGet.path | default "/liveness" }}
```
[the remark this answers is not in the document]
JeffreyThijs: indeed makes more sense, i will fix this

charts/hono/templates/_helpers.tpl
```diff
@@ -479,18 +479,30 @@ The scope passed in is expected to be a dict with keys
...
-    scheme: "HTTP"
-  initialDelaySeconds: {{ default .dot.Values.readinessProbeInitialDelaySeconds .componentConfig.readinessProbeInitialDelaySeconds }}
+    path: {{ $probes.readinessProbe.httpGet.path | default "/readiness" }}
+    port: {{ $probes.readinessProbe.httpGet.port | default "health" }}
+    scheme: {{ $probes.readinessProbe.httpGet.scheme | default "HTTP" }}
+  periodSeconds: {{ $probes.readinessProbe.periodSeconds | default 10 }}
+  failureThreshold: {{ $probes.readinessProbe.failureThreshold | default 3 }}
+  initialDelaySeconds: {{ $probes.readinessProbe.initialDelaySeconds | default 20 }}
+  successThreshold: {{ $probes.readinessProbe.successThreshold | default 1 }}
+  timeoutSeconds: {{ $probes.readinessProbe.timeoutSeconds | default 1 }}
```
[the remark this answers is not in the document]
JeffreyThijs: i just took the k8s defaults, but here we should maybe increase it

charts/hono/values.yaml
```diff
@@ -1823,12 +1794,15 @@ jaegerBackendExample:
...
-  readinessProbeInitialDelaySeconds: 10
+  # probes contains the configuration of the liveness and readiness probes.
+  # If not set the top level probes config will be used
+  probes:
+    livenessProbe:
+      httpGet:
+        path: "/"
+    readinessProbe:
+      httpGet:
+        path: "/"
```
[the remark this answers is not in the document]
JeffreyThijs: they are just using the default ones

charts/hono/templates/artemis/artemis-deployment.yaml
```diff
@@ -65,21 +65,21 @@ spec:
...
-          periodSeconds: 17
+          initialDelaySeconds: {{ $args.componentConfig.probes.livenessProbe.initialDelaySeconds }}
+          periodSeconds: {{ $args.componentConfig.probes.livenessProbe.periodSeconds }}
+          timeoutSeconds: {{ $args.componentConfig.probes.livenessProbe.timeoutSeconds }}
           exec:
             command:
             - "/bin/bash"
-            - {{ printf "%s/liveness-probe.sh" $args.configMountPath | quote }}
-          timeoutSeconds: 3
+            - {{ printf "%s/liveness-probe.sh" $args.configMountPath }}
```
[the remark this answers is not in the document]
JeffreyThijs: deleted this due to a merge conflict i think, i will revert it

charts/hono/values.yaml
```diff
@@ -30,6 +30,12 @@ honoContainerRegistry: "index.docker.io"
 # be added to every deployment and statefulset.
 imagePullSecrets: []
 
+probes:
```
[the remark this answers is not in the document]
JeffreyThijs: > I like the overall approach :-) However, in a service release, we cannot simply remove the existing properties for configuring the probe delays. Instead, we should keep them an mark them as deprecated, providing a hint what to use instead. We can, however, keep them empty and implement logic, that uses the old probes if the values are set explicitly, but otherwise uses your new probes based on the values of the new properties in values.yaml. WDYT?

yeah my bad, i'll make it backwards compatible

charts/hono/values.yaml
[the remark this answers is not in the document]
sophokles73: sure, but before your changes the initial delay for the liveness probe was 300 secs and that for the readiness probe was 10 secs. But now there is no delay at all, if I am not mistaken ...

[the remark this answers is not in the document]
sophokles73: ah, now I see what you meant. The probe properties defined here are merged with the top level probe definitions ...

JeffreyThijs:
@sophokles73, i think i solved all remarks. Could you check again?

charts/hono/values.yaml
```diff
@@ -1663,12 +1709,24 @@ amqpMessagingNetworkExample:
...
+    # probes allows you to overwrite the global probes config values for this specific component
+    probes:
+      livenessProbe:
+        httpGet:
+          path: "/healthz"
+        periodSeconds: 9
+      readinessProbe:
+        httpGet:
+          path: "/healthz"
+        periodSeconds: 5
```
sophokles73, line 1729: IMHO we should set the initialDelaySeconds in the probes here and leave the deprecated properties empty. That way we also provide an example of how to set the values in the desired way ...

charts/hono/values.yaml
```diff
@@ -1740,16 +1798,22 @@ amqpMessagingNetworkExample:
...
-      # Refer to https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
-      # for a description of the properties' semantics.
+      # probes allows you to overwrite the global probes config values for this specific component
+      probes:
+        livenessProbe:
+          periodSeconds: 17
+          timeoutSeconds: 3
+        readinessProbe:
+          periodSeconds: 10
+          timeoutSeconds: 3
```
sophokles73, line 1816: IMHO we should set the initialDelaySeconds in the probes here and leave the deprecated properties empty. That way we also provide an example of how to set the values in the desired way ...

charts/hono/values.yaml
```diff
@@ -1791,12 +1855,30 @@ dataGridExample:
...
+      failureThreshold: 5
+      successThreshold: 1
+      timeoutSeconds: 3
+    readinessProbe:
+      httpGet:
+        path: "/rest/v2/cache-managers/routing-info/health/status"
+        port: "hotrod"
+      failureThreshold: 5
+      successThreshold: 1
+      timeoutSeconds: 3
```
sophokles73, line 1881: IMHO we should set the initialDelaySeconds in the probes here and leave the deprecated properties empty. That way we also provide an example of how to set the values in the desired way ...

charts/hono/values.yaml
```diff
@@ -1823,12 +1905,21 @@ jaegerBackendExample:
...
   # configuration property of the component's readiness probe.
+  # The value of the top level "readinessProbeInitialDelaySeconds" property will be used if not set.
   readinessProbeInitialDelaySeconds: 10
+  probes:
+    livenessProbe:
+      httpGet:
+        path: "/"
+    readinessProbe:
+      httpGet:
+        path: "/"
```
sophokles73, line 1922: IMHO we should set the initialDelaySeconds in the probes here and leave the deprecated properties empty. That way we also provide an example of how to set the values in the desired way ...

sophokles73 (commented):
We are getting close :-)
Can you also please add a corresponding entry to the release notes in the *README.md*?

charts/hono/values.yaml
JeffreyThijs: sure, i'll fix it. I will also adapt that the probe config gets the priority over `livenessProbeInitialDelaySeconds` and `readinessProbeInitialDelaySeconds` as it is now the other way around i think.

sophokles73: Actually, IMHO it is correct that `livenessProbeInitialDelaySeconds` and `readinessProbeInitialDelaySeconds` get priority over the values configured in the `probes` properties. In the end, we want to make sure that users who set these (deprecated) values still get these values applied (instead of the default values that you put into the `probes`). Or am I mistaken?

sophokles73:
@JeffreyThijs can you also amend the README's release notes accordingly?

charts/hono/values.yaml
JeffreyThijs: you are right, i reverted the commit.

charts/hono/README.md
```diff
@@ -116,6 +116,13 @@ The command removes all the Kubernetes components associated with the chart and
   *extraVolumeMounts* properties which allow mounting any kind of volume supported by kubernetes into the container file
   system.
 
+### 2.0.4
+
+* *livenessProbeInitialDelaySeconds* and *readinessProbeInitialDelaySeconds* have been deprecated due to the introduction of the
+  *probes* parameter which allows you to configure the HTTP GET probes more specifically instead of just the initialDelaySeconds. The
+  *probes* parameter is configurable globally but can be overwritten per component.
+
```
sophokles73, line 124: the text is great but can you please move it *before* the 2.0.0 release so that the most recent news are always at the top?

sophokles73 (changes requested):

charts/hono/README.md
JeffreyThijs: updated

charts/hono/values.yaml
```diff
@@ -1791,12 +1859,32 @@ dataGridExample:
...
+      timeoutSeconds: 3
+      initialDelaySeconds: 300
+    readinessProbe:
+      httpGet:
+        path: "/rest/v2/cache-managers/routing-info/health/status"
+        port: "hotrod"
+      failureThreshold: 5
+      successThreshold: 1
+      timeoutSeconds: 3
+      initialDelaySeconds: 20
```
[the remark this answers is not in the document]
JeffreyThijs: updated

sophokles73 (approved):
Awesome! Thanks for contributing :+1:

sophokles73:
@JeffreyThijs your commit message contains a `Co-authored-by` footer instead of the `Signed-off-by` footer. Does this mean that the code in the PR has been authored by others as well?

Issue #391, opened 2022-08-17
Unable to override basic auth credentials for Ditto's nginx


refi64:
Ditto's Helm chart by default places the API behind nginx, secured using the [default `ditto:ditto` credentials](https://github.com/eclipse/packages/blob/14ea942aa5814e44a58de09e3bb94bfc5ca18901/charts/ditto/nginx-config/nginx.htpasswd). Unfortunately, it doesn't seem to be possible to override this file in the [configmap](https://github.com/eclipse/packages/blob/master/charts/ditto/templates/nginx-config.yaml#L26-L27) in any (straightforward?) way, i.e. you can't actually use basic auth without using the hardcoded credentials, and the only alternative is JWTs. (I'm guessing JWTs are the encourages authentication route, but for basic PoC purposes, it's still helpful to be able to use basic auth with custom credentials.)

thjaeckle:
@refi64 thanks for reporting 👍 You're absolutely right - this should be possible to adjust easily when using the Chart.
We'll look into it.

thjaeckle:
@refi64 thanks for reporting, we fixed that by adding a new configuration option to provide additional users, either already hashed via `htpasswd` or in cleartext in the Helm values.

refi64:
That looks great, thanks! I have to ask though, should the ConfigMap actually be a Secret, since it's storing authentication information inside?

thjaeckle:
That is a very good point - thanks for raising that .. we will probably do another PR

stmaute:
I have created another PR https://github.com/eclipse/packages/pull/393 to fix this.

PR #392, opened 2022-08-18
Extend basic auth configuration for Ditto

stmaute:
removed nginx.htpasswd and configure basic auth in another way;
add basicAuthUsers to values.yaml where multiple users can be configured for basic auth;
bump nginx version to 1.23.1;

Fixes: #391

### DIFF
--- a/charts/ditto/values.yaml
+++ b/charts/ditto/values.yaml
@@ -92,9 +92,6 @@ global:
     ditto:
       user: ditto
       password: ditto
-    ditto2:
-      user: ditto2
-      password: ditto2
   ## jwtOnly controls whether only OpenID-Connect authentication is supported
   ## if false, basicAuth is used
   ## ref: https://www.eclipse.org/ditto/installation-operating.html#openid-connect
--- a/charts/ditto/values.yaml
+++ b/charts/ditto/values.yaml
@@ -86,12 +86,20 @@ openshift:
 ## ----------------------------------------------------------------------------
 ## global configuration shared by all components
 global:
+  # Nginx Basic Auth config
+  # either configure one or more basic auth user in the basicAuthUsers section
+  # where you can specify username and password in clear text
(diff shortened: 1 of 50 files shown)


charts/ditto/values.yaml
```diff
@@ -85,6 +86,12 @@ openshift:
 ## ----------------------------------------------------------------------------
 ## global configuration shared by all components
 global:
+  # default basic auth user used for authentication in nginx is ditto with password ditto
+  # add more users to the basicAuthUsers array or replace the default user
+  basicAuthUsers:
+    ditto:
+      user: ditto
+      password: ditto
```
thjaeckle, line 94: we should add an option to also be able to provide the already hashed user/password  list here, so that you don't have to put cleartext passwords in your Helm config

proposal: 
* support 2 different types of username/password combinations to configure:
   * `basicAuthUsers` (e.g. being the cleartext password like you already configured it)
   * `hashedBasicAuthUsers` (a list of already by `htpasswd` hashed username/password lines)

thjaeckle (changes requested):
Thanks @stmaute for fixing that so quickly.
I have one concern that with the approach now you have to put your cleartext password in the Helm chart's config.
Could we also support configuring a list of `htpasswd` hashed entries?

charts/ditto/values.yaml
stmaute: yes sure this is a good idea. I can add the `hashedBasicAuthUsers` type. but then I suggest that it is only possible to configure one of the types. because otherwise it will get really complicated. @thjaeckle wdyt?

thjaeckle: yes, that should be sufficient - either one or the other 👍

stmaute: @thjaeckle I have added a hashedBasicAuthUsers section and now it is possible to use both approaches to configure basic auth users.

thjaeckle (approved):
Changes LGTM - thx @stmaute 👍

PR #393, opened 2022-08-23
Use Secret for nginx authentication

stmaute:
switched from kind ConfigMap to Secret for nginx-auth;
adapted nginx-deployment.yaml to use the secret instead of the configmap;


thjaeckle (approved):
LGTM :+1:

PR #395, opened 2022-08-30
Enable self signed certs for Ditto managed Kafka connection in c2e package

kalinkostashki:
This will enable configuration/support of self-signed certificates for the Kafka connection of Ditto.


packages/cloud2edge/requirements.yaml
```diff
@@ -12,8 +12,8 @@
 #
 dependencies:
   - name: hono
-    version: ^1.10.18
+    version: ~2.0.7
     repository: "https://eclipse.org/packages/charts/"
   - name: ditto
-    version: ~2.4.0
+    version: ~2.4.1
```
thjaeckle, line 18: @kalinkostashki this must be the chart version of Ditto (you used the "appVersion") - the most recent chart version is 2.5.5 (including Ditto 2.4.1)

kalinkostashki: done

packages/cloud2edge/Chart.yaml
```diff
@@ -11,8 +11,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.2.4
-appVersion: 0.2.4
+version: 0.3.0
+appVersion: 0.3.0
```
thjaeckle, line 15: @kalinkostashki I would suggest to only increase to `0.2.5` here - as this can be seen as "bugfix" of the chart.
When using new Ditto/Hono major versions we can update to `0.3.0`

packages/cloud2edge/requirements.yaml
```diff
@@ -12,8 +12,8 @@
 #
 dependencies:
   - name: hono
-    version: ^1.10.18
+    version: ~2.0.7
```
thjaeckle, line 15: @kalinkostashki the version range was before starting with `^` - I think we should keep it and not change to `~`

kalinkostashki: ok, will fix this.

packages/cloud2edge/templates/post-install-job.yaml
```diff
@@ -56,7 +57,13 @@ spec:
           {{- end }}
 
           {{- if ( has "kafka" .Values.hono.messagingNetworkTypes ) }}
-          curl -X POST -u devops:$PWD -H 'Content-Type: application/json' --data-binary @/var/run/c2e/post-install-data/hono-kafka-connection.json
+          cp /var/run/c2e/post-install-data/insert-kafka-certs.sh $curlHome/insert-kafka-certs.sh
+
+          chmod +x $curlHome/insert-kafka-certs.sh
+
+          sh $curlHome/insert-kafka-certs.sh
```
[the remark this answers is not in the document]
kalinkostashki: you're right - will fix this.

packages/cloud2edge/values.yaml
```diff
@@ -23,9 +23,14 @@ honoConnection:
 # Configuration properties for Eclipse Hono.
 hono:
 
+# Explicitly check for kafkaMessagingClusterExample because the post-install-job needs to know if it should load
+# the certificates from the kubernetes secrets and attach them to the hono-kafka-connection.json.
+  kafkaMessagingClusterExample:
+    enabled: true
```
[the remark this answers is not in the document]
kalinkostashki: I was under the incorrect impression that I wouldn't have access to it otherwise inside the post-install-job.yaml otherwise.
Will fix :)

packages/cloud2edge/post-install/insert-kafka-certs.sh
```diff
@@ -0,0 +1,15 @@
...
+cp /var/run/c2e/post-install-data/hono-kafka-connection.json $curlHome/hono-kafka-connection.json
+chmod 777 $curlHome/hono-kafka-connection.json
+
+jsonPath=$curlHome/hono-kafka-connection.json
+
+if [ -e $certPath ]
+  then
+    cert="$(cat /var/run/c2e/kafka-connection-cert/tls.crt | tr -d '\n' | sed 's/E-----/E-----\\\\n/g' | sed 's/-----END/\\\\n-----END/g')"
+    sed -i '/uri/ a '"\"ca\": \"$cert\","'' $jsonPath
+fi
```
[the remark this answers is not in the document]
kalinkostashki: I will added the necessary comments here.

packages/cloud2edge/requirements.yaml
calohmn, line 15: The chart changes for Hono 2.0 included renaming some K8s resources. We have dropped the `-vertx` suffix there.
An HTTP protocol adapter pod name now contains just `adapter-http` instead of `adapter-http-vertx`.

That means that in all the cloud2edge package files, in fact also in the homepage `prereqs.md`, all occurrences of `adapter-[*]-vertx` need to be replaced with `adapter-[*]`.

An example here is the 
`getService adapter-amqp-vertx "amqp amqps" AMQP_ADAPTER` 
line in `setCloud2EdgeEnv.sh`.

packages/cloud2edge/Chart.yaml
```diff
@@ -11,8 +11,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.3.0
-appVersion: 0.3.0
+version: 0.2.5
```
sophokles73, line 14: the version number should be *increased*, not *decreased* ...

sophokles73 (changes requested):
The credentials defined in values.yaml and which are used for Ditto's connection to the Kafka broker need to be added to the Kafka broker's configuration properties. Otherwise, the connection will never succeed. The relevant properties in the Hono chart's values.yaml are `kafka.auth.sasl.jaas.clientUsers` and `kafka.auth.sasl.jaas.clientPasswords` ...

packages/cloud2edge/Chart.yaml
kalinkostashki: this is an increase.
I had suggested 0.3.0 from 0.2.4 but @thjaeckle suggested that 0.2.5 is more appropriate.

sophokles73: my bad, I only saw the diff to the previous commit ;-)

calohmn: Seeing that the changes here involve the upgrade from Hono 1.x to Hono 2.1, I think 0.3.0 would be more appropriate.

packages/cloud2edge/requirements.yaml
```diff
@@ -12,8 +12,8 @@
...
     repository: "https://eclipse.org/packages/charts/"
   - name: ditto
-    version: ~2.4.0
+    version: ~2.5.5
```
calohmn, line 18: It would be good to increase this to 2.5.6, including the fix for #398.

calohmn:
> @kalinkostashki in the meantime we could also reference the just merged Hono 2.1.0 chart

There is Hono chart version 2.1.1 now.

calohmn:
@kalinkostashki For the Hono 2.x update, there are further changes needed:
In `values.yaml`, the
```
    extraSecretMounts:
      permissions:
        secretName: "permissions"
        mountPath: "/var/run/hono/auth"
```
block needs to be replaced with
```
    extraVolumes:
      - name: "permissions"
        secret:
          secretName: "permissions"
    extraVolumeMounts:
      - name: "permissions"
        mountPath: "/var/run/hono/auth"
```
and in `permissions-secret.yaml` the metadata-name can then be set to `permissions` as fixed value.

calohmn:
@kalinkostashki Could you do a rebase here? There has been an update to the cloud2edge chart (#406).

packages/cloud2edge/requirements.yaml
kalinkostashki: done :)

packages/cloud2edge/templates/NOTES.txt
```diff
@@ -13,9 +13,9 @@ You can check the current status by issuing the following command:
 Once all services are up and running, the output should look similar to this:
 
 NAME                                       {{ "READY   STATUS    RESTARTS   AGE" | indent ( len .Release.Name ) }}
-{{ .Release.Name }}-adapter-amqp-vertx-65cfb4d675-g5wn4       1/1     Running   0          3m15s
-{{ .Release.Name }}-adapter-http-vertx-66bd6bb89c-mng5t       1/1     Running   0          3m15s
-{{ .Release.Name }}-adapter-mqtt-vertx-765fcd578b-5rl7n       1/1     Running   0          3m15s
+{{ .Release.Name }}-adapter-amqp-65cfb4d675-g5wn4       1/1     Running   0          3m15s
+{{ .Release.Name }}-adapter-http-66bd6bb89c-mng5t       1/1     Running   0          3m15s
+{{ .Release.Name }}-adapter-mqtt-765fcd578b-5rl7n       1/1     Running   0          3m15s
```
sophokles73, line 18: IMHO this breaks the indentation of the *STATUS* column, i.e. you will need to add *space* characters for the removed `vertx-` suffix ...

sophokles73 (commented):
the Hono connection credentials configured in the *honoConnection* property need to also be added to the kafka broker configuration, i.e. the *hono.kafka.

kalinkostashki:
> @kalinkostashki in the meantime we could also reference the just merged Hono 2.1.0 chart

done

packages/cloud2edge/templates/NOTES.txt
kalinkostashki: sorry about the status - it seems that when I tested I was blind :D 
also added the respective properties(kafka.auth.sasl.jaas.clientUsers and kafka.auth.sasl.jaas.clientPasswords) to the values.yaml - is this sufficient or do you propose/prefer a different approach ?

packages/cloud2edge/requirements.yaml
kalinkostashki: done

packages/cloud2edge/requirements.yaml
```diff
@@ -12,8 +12,8 @@
 #
 dependencies:
   - name: hono
-    version: ^1.10.18
+    version: ~2.1.0
```
sophokles73, line 15: Can we change that to `^2.1.0` to also allow future minor versions?

packages/cloud2edge/values.yaml
```diff
@@ -23,9 +23,18 @@ honoConnection:
...
 hono:
 
+  kafka:
+    auth:
+      sasl:
+        jaas:
+          clientUsers:
+            - "ditto-c2e"
+          clientPasswords:
+            - "verysecret"
```
sophokles73, line 33: I am not sure if this simply *overwrites* the credentials defined in the Hono chart of if it *adds* the credentials to the ones defined in the Hono chart. In any case, we need to make sure that the credentials from the Hono chart are being kept as well ...

sophokles73 (changes requested):

packages/cloud2edge/Chart.yaml
kalinkostashki: updated to 0.3.0

packages/cloud2edge/values.yaml
kalinkostashki: I will check in detail how helm behaves in this case.
It seems though on a first look of the kafka logs that hono is accessing it:
`INFO Created log for partition hono.command_internal.HonoAMQPAdapter-e7104df7-34b9-408c-b158-ba7f89fb44cd-0 in /bitnami/kafka/data/hono.command_internal.HonoAMQPAdapter-e7104df7-34b9-408c-b158-ba7f89fb44cd-0 with properties {} (kafka.log.LogManager)`
But like I said - will look a bit deeper.

kalinkostashki: checked and indeed it overwrites the values from hono.
So either I put both in values of cloud2edge or find another way to do this.

sophokles73: > ... it works because hono takes the first values

Yes, then I guess we are all good. As long as these credentials match the ones set in `honoConnection` we should be fine ...

sophokles73: FMPOV we should keep both sets of credentials so that the Hono command line client can still successfully connect to the Kafka broker. The `ditto-c2e/verysecret` credentials should be defined second in order not to break things ...

packages/cloud2edge/requirements.yaml
kalinkostashki: done

packages/cloud2edge/values.yaml
kalinkostashki: ok, I've added them like requested also tested a deployment and confirmed that hono-secret is indeed the first one in the secret.

sophokles73 (commented):
LGTM

now you only need to rebase ...

sophokles73 (approved):

sophokles73:
@calohmn anything else to add or can we merge this PR?

calohmn:
Can be merged FMPOV. I would just suggest to squash the commits before the merge.

PR #397, opened 2022-08-30
[#396] Use Hono 2.1.0 in Hono chart

### LINKED ISSUE #396: Use Hono 2.1.0 in Hono chart

sophokles73:
Changed configuration so that components now dynamically retrieve
validation key(s) from Auth server instead of requiring static
configuration of Auth server's public key.

Fixes #396

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,17 +15,18 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 2.0.7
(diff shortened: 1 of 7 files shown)


charts/hono/Chart.yaml
```diff
@@ -15,14 +15,15 @@ name: hono
...
+appVersion: 2.1.0
 keywords:
 - iot-chart
 - IoT
 - connectivity
 - messaging
+- Kafka
 home: https://www.eclipse.org/hono/
 sources:
 - https://github.com/eclipse/hono
```
calohmn, line 29: This PR could be used to update the location here.

calohmn (approved):
LGTM

PR #374, opened 2022-06-03
Update mongodb chart version to 11.2.0

stmaute:
use mongodb version 4.4.13;
update mongodb config to disable authentication;

PR #403, opened 2022-09-09
Fix high CPU usage of Ditto mongodb container

calohmn:
When testing the cloud2edge Helm chart version with the latest Ditto chart version (using the changes from #395), I noticed that the `ditto-mongodb` pod never got ready. There were always liveness/readiness check timeouts.

The issue here seems to be, that with the MongoDB chart version 11.2.0 (see #374), there is a considerably higher consumption of CPU and also memory resources due to the usage of the MongoDB `mongosh` tool instead of `mongo` for the health probes.

This is described and tracked in https://github.com/bitnami/charts/issues/10316, https://github.com/bitnami/charts/issues/10264 and [MONGOSH-1240](https://jira.mongodb.org/browse/MONGOSH-1240).

The last chart version that doesn't show this issue is 11.1.10. It uses the same MongoDB version as the 11.2.0 chart (MongoDB 4.4.13). Therefore I think the straightforward solution for now is a downgrade to 11.1.10.

----
Here some charts showing metrics after deploying Ditto with mongoDB chart 11.2.0:
MongoDB_11 2 0

And here with 11.1.10:
MongoDB_11 1 10

### DIFF
--- a/charts/ditto/Chart.yaml
+++ b/charts/ditto/Chart.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
+# Copyright (c) 2019, 2022 Contributors to the Eclipse Foundation
 #
 # See the NOTICE file(s) distributed with this work for additional
 # information regarding copyright ownership.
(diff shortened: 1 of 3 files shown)


stmaute (approved):
LGTM 👍🏻

thjaeckle (approved):
LGTM, thx @calohmn 👍

PR #406, opened 2022-09-09
Fix c2e's Kafka setup when using an external cluster

refi64:
The credentials were always assumed to be those for the example one.

### DIFF
--- a/packages/cloud2edge/post-install/hono-kafka-connection.json
+++ b/packages/cloud2edge/post-install/hono-kafka-connection.json
@@ -10,7 +10,13 @@
       "name": "eclipse-hono-via-kafka",
       "connectionType": "kafka",
       "connectionStatus": "open",
+      {{- if .Values.hono.kafkaMessagingClusterExample.enabled }}
       "uri": "tcp://{{ .Values.honoConnection.username }}:{{ .Values.honoConnection.password }}@{{ .Release.Name }}-kafka:9092",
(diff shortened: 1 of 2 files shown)


calohmn:
@refi64 Thanks for the PR. In general, chart changes have to be accompanied with an increase of the chart version in the corresponding `Chart.yaml`. In the case of the cloud2edge chart that also applies to the `appVersion` (setting both to `0.2.5` then).

packages/cloud2edge/post-install/hono-kafka-connection.json
```diff
@@ -151,8 +157,14 @@
         }
       ],
       "specificConfig": {
+        {{- if .Values.hono.kafkaMessagingClusterExample.enabled }}
         "saslMechanism": "plain",
         "bootstrapServers": "{{ .Release.Name }}-kafka:9092",
+        {{- else }}
+        {{ $kafka := .Values.hono.adapters.kafkaMessagingSpec.commonClientConfig }}
+        "saslMechanism": "{{ (index $kafka "sasl.mechanism") | lower }}",
```
calohmn, line 165: Could you use "plain" as fallback here? The "sasl.mechanism" property might not have been set in `commonClientConfig`.

refi64:
@calohmn oops, version updated!

calohmn (approved):
LGTM

Thanks for contributing :+1:

PR #407, opened 2022-09-12
Use Jaeger all-in-one OTLP endpoint

calohmn:
Enable the Jaeger all-in-one container OTLP endpoint and send traces directly to that endpoint, skipping collector sidecar deployment in that case. The sidecars are still deployed if Jaeger example backend deployment is disabled and a `otelCollectorAgentConfigMap` is defined.

This reduces the overall Hono deployment memory footprint when using the Jaeger example backend and enables using the Jaeger OTLP endpoint also from other components deployed with Hono (e.g. Ditto).


sophokles73 (approved):
LGTM

PR #411, opened 2022-09-14
[#410] Do not set node port on Services when using LoadBalancers

### LINKED ISSUE #410: NodePort is defined even if LoadBalancer is used
The Hono Service definitions include a *NodePort* definition when the *useLoadBalancer* property is set to `true` but the *serviceType* property is unset. However, in this case, the Service definitions should not contain a NodePort.

sophokles73:
The Hono service definitions had always also defined a node port unless the serviceType property had been set to "ClusterIP" explicitly.

This has been changed so that the node port is now only defined if the
 serviceType property is explicitly set to "NodePort" or if the
useLoadBalancer property is set to "false".

Fixes #410


calohmn (approved):
LGTM

PR #415, opened 2022-09-22
Support creating the default c2e tenant w/o the demo device

refi64:
This is useful for having a default tenant available for testing purposes without also having a demo device & credentials registered.

### DIFF
--- a/packages/cloud2edge/values.yaml
+++ b/packages/cloud2edge/values.yaml
@@ -11,8 +11,9 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
+tenant: "org.eclipse.packages.c2e"
 demoDevice:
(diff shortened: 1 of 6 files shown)


packages/cloud2edge/values.yaml
```diff
@@ -11,8 +11,9 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
+tenant: "org.eclipse.packages.c2e"
 demoDevice:
-  tenant: "org.eclipse.packages.c2e"
+  enabled: true
```
calohmn, line 16: Since there is also an `enabled` field in Hono device data, I think it would be better to choose a different property name here (e.g. `skip` or `omit`).

Issue #252, opened 2021-06-02
[Hono] Fix service names in example certificates


b-abel:
The configuration for creating example certificates defines _Subject alternative names_ for the service names, like eg. "hono-service-auth.hono". But the service names are prefixed during deployment with the _release name_. It is therefore impossible to include the release name into the certificates. 
All client configuration I found disables hostname verification, so there is no real consequence during runtime. But FMPOV the configuration is misleading and does not serve as a good example. 

The easiest solution would probably be to remove the release name prefix from the service names. The impact should be small because in practice probably no one deploys Hono multiple times into the same Kubernetes namespace.

Another solution could be to define a "standard release name" and create the certificates for this name. This could be accompanied by informing the user via the `NOTES.txt` about the release name to be used, or even by a check that requires the release name.

b-abel:
FMPOV it would be great if hostname verification could be enabled at least for the internal connections (i.e. between the components running in a Kubernetes cluster in the same namespace).

sophokles73:
I agree that this is not optimal and I also agree that we could assume (or better require) that there is only one Hono installation per kubernetes namespace. That would allow us to remove the release name from the service names (and all other artifacts that we deploy to the namespace).

b-abel:
> we could assume (or better require) that there is only one Hono installation per kubernetes namespace

Removing the release name from the names of services, deployments, etc. would lead to an error if you try to deploy Hono a second time to a namespace. FMPOV this would be sufficient. Or are you thinking of additional explicit checking?

PR #416, opened 2022-09-26
allow overriding release name

vhdirk:
Fixes https://github.com/eclipse/packages/issues/430

Signed-off-by: Dirk Van Haerenborgh <dirk.vanhaerenborgh@aloxy.io>

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 2.2.0
+version: 2.2.1
 # Version of Hono being deployed by the chart
 appVersion: 2.2.0
 keywords:
--- a/charts/hono/templates/_helpers.tpl
+++ b/charts/hono/templates/_helpers.tpl
@@ -240,7 +240,7 @@ messaging:
 {{- if .dot.Values.amqpMessagingNetworkExample.enabled }}
   name: {{ printf "Hono %s" .component | quote }}
   amqpHostname: "hono-internal"
-  host: {{ printf "%s-dispatch-router" ( include "hono.fullname" .dot ) | quote }}
+  host: {{ printf "%s-dispatch-router" ( include "hono.fullname" . ) | quote }}
   port: 5673
   keyPath: {{ .dot.Values.adapters.amqpMessagingNetworkSpec.keyPath | quote }}
   certPath: {{ .dot.Values.adapters.amqpMessagingNetworkSpec.certPath | quote }}
--- a/charts/hono/templates/_helpers.tpl
+++ b/charts/hono/templates/_helpers.tpl
@@ -102,7 +102,7 @@ The scope passed in is expected to be a dict with keys
 - "component": the value to use for the "app.kubernetes.io/component" label
 */}}
 {{- define "hono.metadata" -}}
-name: {{ printf "%s-%s" .dot.Release.Name .name | quote }}
+name: {{ printf "%s-%s" (include "hono.fullname" .dot ) .name | quote }}
 namespace: {{ .dot.Release.Namespace | quote }}
 labels:
   app.kubernetes.io/name: {{ include "hono.name" .dot | quote }}
@@ -237,7 +237,7 @@ messaging:
 {{- if .dot.Values.amqpMessagingNetworkExample.enabled }}
   name: {{ printf "Hono %s" .component | quote }}
   amqpHostname: "hono-internal"
-  host: {{ printf "%s-dispatch-router" .dot.Release.Name | quote }}
+  host: {{ printf "%s-dispatch-router" ( include "hono.fullname" .dot ) | quote }}
   port: 5673
   keyPath: {{ .dot.Values.adapters.amqpMessagingNetworkSpec.keyPath | quote }}
   certPath: {{ .dot.Values.adapters.amqpMessagingNetworkSpec.certPath | quote }}
@@ -263,7 +263,7 @@ The scope passed in is expected to be a dict with keys
 kafka:
 {{- if .dot.Values.kafkaMessagingClusterExample.enabled }}
   commonClientConfig:
-    {{- $bootstrapServers := printf "%[1]s-%[2]s-0.%[1]s-%[2]s-headless:%d" .dot.Release.Name .dot.Values.kafka.nameOverride ( .dot.Values.kafka.service.ports.client | int ) }}
+    {{- $bootstrapServers := printf "%[1]s-%[2]s-0.%[1]s-%[2]s-headless:%d" ( include "hono.fullname" .dot ) .dot.Values.kafka.nameOverride ( .dot.Values.kafka.service.ports.client | int ) }}
     bootstrap.servers: {{ $bootstrapServers | quote }}
   {{- if eq .dot.Values.kafka.auth.clientProtocol "sasl_tls" }}
     security.protocol: "SASL_SSL"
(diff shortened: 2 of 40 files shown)


calohmn:
@vhdirk The goal of #252 is for the _subject alternative names_ of the example certificates to always match the Hono service names, so that we can enable hostname verification in the clients. With the release name being part of the service name, matching names cannot be guaranteed. Therefore the discussion in #252 lead to the plan to remove the release name from the Hono service names, meaning service names are fixed (see https://github.com/eclipse/packages/issues/252#issuecomment-883204654). 

In your PR here, this approach is not taken. Can you explain what the relation to #252 is?

Just looking at the PR changes itself, I see that these would harmonize K8s resource naming with the approach taken in the ditto chart. This would mean that in the cloud2edge chart, service names would then be
```
c2e-hono-adapter-amqp
c2e-hono-service-auth
c2e-ditto-gateway
c2e-ditto-nginx
```
and not 
```
c2e-adapter-amqp
c2e-service-auth
c2e-ditto-gateway
c2e-ditto-nginx
```
any more, making the naming clearer. This will need changes in the cloud2edge chart as well though (e.g. in scripts invoking the services). Therefore we have to think about versioning and compatibility here. (To stay compatible with the cloud2edge chart, we would even have to make this change here a major version jump. I guess we could consider adapting the cloud2edge chart Hono requirements version range though (to ~2.1.0), requiring only a minor version jump here.)

vhdirk:
When asking on gitter if there was an option to override the release name, I was pointed to #252, which is why I referenced it.
I chose not to remove the release name entirely, since we internally also have a service called 'adapter'. The removal of release names from Hono would make it too ambiguous. And we also like the parity with ditto.

I can't imagine removing the prefix altogether would be the only option with regard to hostname verification?

sophokles73:
> When asking on gitter if there was an option to override the release name, I was pointed to https://github.com/eclipse/packages/issues/252,

I am sorry, that was my fault. I had simply thought that we could kill two birds with one stone. However, I guess this PR has value on its own, even if it does not address the host name verification problem itself. I agree with @calohmn that we need be transparent about backward compatibility (or the lack thereof) by means of choosing a proper version number increase.

@calohmn apart from the version and corresponding release notes entry, is there anything else that needs to be changed?

> I can't imagine removing the prefix altogether would be the only option with regard to hostname verification?

Probably not, I am currently thinking in the direction to (dynamically) create the demo certificates by means of cert-manager, i.e. making cert-manager a pre-requisite for the Hono chart when using demo certs.

charts/hono/templates/hono-service-device-registry-base/hono-service-device-registry-post-install-job.yaml
```diff
@@ -48,6 +48,6 @@ spec:
       volumes:
       - name: post-install-data
         configMap:
-          name: {{ printf "%s-post-install-device-registry-conf" .Release.Name }}
+          name: {{ include "hono.fullname" . }}"-post-install-device-registry-conf"
```
sophokles73, line 51: this produces `"hono\"-post-install-device-registry-conf\"` which causes the deployment test to fail

I suggest to use
```yaml
name: {{ printf "%s-post-install-device-registry-conf" ( include "hono.fullname" . ) }}
```

sophokles73 (changes requested):

charts/hono/templates/hono-service-device-registry-base/hono-service-device-registry-post-install-job.yaml
vhdirk: fixed it for al occurrences of the same thing

charts/hono/Chart.yaml
```diff
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 2.1.5
```
calohmn, line 18: Regarding cloud2edge compatibility, we have 2 options here:
- use version `3.0.0` here. That means the current  cloud2edge chart won't reference it (`^2.1.0` used there now).
- use version `2.2.0` here and first create a new cloud2edge chart version that only changes the Hono dependency requirement from `^2.1.0` to `~2.1.0`.

I would tend towards the 2nd option here. Having just the patch-level version range in the cloud2edge chart offers more flexibility regarding Hono chart changes affecting the cloud2edge chart, without having to do a major version jump of the Hono chart each time.
@sophokles73 WDYT?

charts/hono/Chart.yaml
```diff
@@ -15,7 +15,7 @@ name: hono
...
+version: 2.1.6
```
calohmn, line 18: @vhdirk Can you also add an entry in the `Release Notes` section of the `README.md`?

sophokles73 (changes requested):

charts/hono/templates/_helpers.tpl
```diff
@@ -102,7 +102,7 @@ The scope passed in is expected to be a dict with keys
 - "component": the value to use for the "app.kubernetes.io/component" label
 */}}
 {{- define "hono.metadata" -}}
-name: {{ printf "%s-%s" .dot.Release.Name .name | quote }}
+name: {{ printf "%s-%s" (include "hono.fullname" .dot ) .name | quote }}
```
calohmn, line 105: It looks like there are still some more replacements like this needed:
- the pod list in `NOTES.txt`
- data-grid server name in `hono-data-grid.xml`
- host entries in `qdrouterd.json`
- `HTTP_BASE_URL` in `add_example_data_device_registry.sh`
- `{{ .Release.Name }}-kafka-example-keys` in `values.yaml`

charts/hono/templates/example-data-grid/statefulset.yaml
```diff
@@ -74,5 +74,5 @@ spec:
       volumes:
       - name: conf
         configMap:
-          name: {{ printf "%s-%s-conf" .Release.Name $args.name }}
+          name: {{ printf "%s-%s-conf" (include "hono.name" .) $args.name }}
```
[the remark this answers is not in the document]
vhdirk: Not really, I took inspiration from the ditto helm chart, that is all. I'll change all occurrences

charts/hono/Chart.yaml
vhdirk: I also gravitate more towared the second option here.

charts/hono/Chart.yaml
vhdirk: done. Changed to v2.2.0 as per https://github.com/eclipse/packages/pull/416#discussion_r1001436531

charts/hono/templates/_helpers.tpl
vhdirk: done

charts/hono/templates/example-data-grid/statefulset.yaml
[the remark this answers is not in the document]
vhdirk: fixed

sophokles73:
@vhdirk with all the discussion around whether this PR actually helps with fixing the SAN verification problem, I have lost track of what this PR actually tries to achieve. Would you mind creating a dedicated issue describing the lacking functionality and/or problems with the currently implemented naming strategy? We can then better assess, if your PR goes in the right direction or should better take another route to achieving the goal.
Sorry for the inconvenience but I guess the discussion will become more focused this way.

vhdirk:
@sophokles73 no problem, that's perfectly understandable. I've created https://github.com/eclipse/packages/issues/430 and linked this PR to it

sophokles73:
@vhdirk can we close this PR?

calohmn:
Closing this PR as this feature was merged via #469.

Issue #417, opened 2022-09-27
Update Ditto chart to use Ditto 3.0 release


thjaeckle:
Ditto 3.0 release is planned for 28/09/2022: https://projects.eclipse.org/projects/iot.ditto/releases/3.0.0

kalinkostashki:
will take care of it when we have the ditto 3.0.0 release.

PR #419, opened 2022-09-28
update ditto chart to 3.0.0

kalinkostashki:
- removed all concierge entries as per its removal as one of the main changes in ditto 3.0.0

Fixes: #417


thjaeckle (approved):
LGTM 👍 

Just needed to adapt how system-properties are passed to the Ditto processed which changed in Ditto 3.0

PR #420, opened 2022-09-28
update cloud2edge chart with ditto 3.0.0

kalinkostashki:
- removed all concierge mentions as per the ditto 3.0.0 introduction
 - increased the respective versions in Chart.yaml and requirements.yaml

Signed-off-by: Kalin Kostashki <kalin.kostashki@bosch.io>

### DIFF
--- a/packages/telemetry-e2e/values.yaml
+++ b/packages/telemetry-e2e/values.yaml
@@ -116,14 +116,6 @@ drogueCloud:
 ditto:
   fullnameOverride: ditto
 
-  concierge:
-    resources:
-      requests:
-        cpu: "100m"
-        memory: 256Mi
-      limits:
-        cpu: "1"
-        memory: 384Mi
   connectivity:
     resources:
       requests:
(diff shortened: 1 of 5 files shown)


packages/telemetry-e2e/values.yaml
```diff
@@ -116,14 +116,6 @@ drogueCloud:
...
   fullnameOverride: ditto
 
-  concierge:
-    resources:
-      requests:
-        cpu: "100m"
-        memory: 256Mi
-      limits:
-        cpu: "1"
-        memory: 384Mi
```
thjaeckle, line 126: @kalinkostashki the "telemetry-e2e" package should not be affected by this PR.
And update of the Ditto version in this package would have to be tracked separately.

kalinkostashki: done

thjaeckle (approved):
LGTM 👍 
thx @kalinkostashki

PR #428, opened 2022-10-18
Allow prometheus path to be configurable via values

jokraehe:
Added a global configuration parameter for the prometheus path.


stmaute (approved):
LGTM 👍🏻

PR #437, opened 2022-12-17
[Hono] Update Bitnami Kafka chart to version 20.x

sophokles73:
Also configured Kafka and Zookeeper JVMs to use at most 512M memory.

### DIFF
--- a/charts/hono/values.yaml
+++ b/charts/hono/values.yaml
@@ -1718,15 +1718,31 @@ kafka:
     # Note that this could require creating RBAC rules, for more information refer to
     # https://github.com/bitnami/charts/tree/master/bitnami/kafka#accessing-kafka-brokers-from-outside-the-cluster
     create: true
+  # limit Kafka JVM's memory consumption
+  heapOpts: "-Xms620M -Xmx620M"
+  resources:
+    requests:
+      memory: "800Mi"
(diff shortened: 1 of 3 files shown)


sophokles73:
closed in favor of #438

charts/hono/values.yaml
```diff
@@ -1718,15 +1718,31 @@ kafka:
...
+        enabled: true
+        clientUser: "zookeeperUser"
+        clientPassword: "zookeeperPassword"
+        serverUsers: "zookeeperUser"
+        serverPasswords: "zookeeperPassword"
+    # limit Zookeeper JVM's memory consumption
+    heapSize: 400
+    resources:
+      requests:
+        memory: "500Mi"
```
calohmn, line 1743: How about keeping the zookeeper default of `256Mi` here for the `requests` value? I would also consider reducing the corresponding kafka resources/requests/memory configuration. Both would reduce the overall minimal memory requirements which could for example be relevant when installing the cloud2edge chart locally for testing purposes.

sophokles73: I had replaced this PR with #438 ...

calohmn: Sorry, mixed up the 2 PRs.

PR #438, opened 2022-12-20
[Hono] Update to latest Bitnami chart versions

sophokles73:
Updated Bitnami Kafka chart to version 20.x.
Updated Bitnami MongoDB chart to version 13.x.

Also configured upper limit for Kafka and Zookeeper JVMs memory consumption.

### DIFF
--- a/charts/hono/values.yaml
+++ b/charts/hono/values.yaml
@@ -1360,6 +1360,13 @@ deviceRegistryExample:
 mongodb:
   # createInstance indicates whether a MongoDB database instance should be created.
   createInstance: false
+  # do not use the default name ("mongodb") because it will lead to a config
+  # property name clash in the device registry pod when using "hono" as the release name
+  nameOverride: "monogodb-server"
+  # run single node instance
+  architecture: "standalone"
+  # Use StatefulSet instead of Deployment when deploying standalone
+  useStatefulSet: false
   # If authentication to be enabled or not.
   # ref: https://docs.mongodb.com/manual/tutorial/enable-authentication/
   auth:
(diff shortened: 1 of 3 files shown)


sophokles73:

charts/hono/values.yaml
```diff
@@ -1718,15 +1715,31 @@ kafka:
...
+        enabled: true
+        clientUser: "zookeeperUser"
+        clientPassword: "zookeeperPassword"
+        serverUsers: "zookeeperUser"
+        serverPasswords: "zookeeperPassword"
+    # limit Zookeeper JVM's memory consumption
+    heapSize: 400
+    resources:
+      requests:
+        memory: "500Mi"
```
calohmn, line 1740: How about keeping the zookeeper default of `256Mi` here for the `requests` value? I would also consider reducing the corresponding kafka resources/requests/memory configuration. Both would reduce the overall minimal memory requirements which could for example be relevant when installing the cloud2edge chart locally for testing purposes.

sophokles73: I am not sure if we can reduce the Java heap size of the Kafka broker even further. The default is 1024M and based on my experience with trying to reduce the heap size when running the ITs locally, I do not believe that we can go down much more.
We might be able to configure a smaller amount of requested memory but IMHO the process(es) will very quickly try to claim more than the initial amount of memory so in the end I do not think that this will help with running at a reduced memory footprint. Have you made other experiences?

calohmn: In my tests, the Kafka pod took 570MiB minimum (having sent some messages for 1 device). So, I agree, there's not much room for reducing the requested memory size (I guess using 650MiB would also work, but then there's not much of a difference to the 800MiB size).
For Zookeeper though, I only saw usage of 130MiB, so I guess using 256MiB as requested size seems feasible there.

calohmn (approved):
LGTM

PR #439, opened 2022-12-20
Ditto chart improvements

thjaeckle:
* migrate to Helm "apiVersion: v2", moving dependencies to Chart.yaml
* get rid of "fullnameOverride" to "ditto-mongodb", instead also use default naming pattern with release name and use placeholder mechanism for determining effective Mongo hostname
* improved nginx config, allowing both nginx based and OpenID based authentication by default
* updated MongoDB to version 5.x


charts/ditto/values.yaml
```diff
@@ -709,7 +709,7 @@ thingsSearch:
 mongodb:
   ## enabled controls usage of dependency chart
   enabled: true
-  fullnameOverride: ditto-mongodb
+  # fullnameOverride: ditto-mongodb
```
kalinkostashki, line 712: Shouldn't this be called "fullname" so it corresponds to the _helpers.tpl and mongodb-secret.yaml variables?

thjaeckle: No, the `_helpers.tpl` uses `.Values.fullnameOverride`.

kalinkostashki (approved):
Looks good to me :)

PR #441, opened 2023-01-09
updated ditto to 3.1.1

kalinkostashki:
- respectively bumped up cloud2edge package to use ~3.1.0 charts


thjaeckle (approved):
LGTM, thx @kalinkostashki

PR #446, opened 2023-02-09
update ditto to 3.1.2

JeffreyThijs:


thjaeckle (approved):
Thanks for the update @JeffreyThijs 
LGTM 👍

PR #451, opened 2023-03-05
Allow to define existing secret with Ditto gateway passwords 

kyberpunk:
Hello. This pull request is adding option to override Ditto gateway secret by custom secret created outside of Ditto helm chart. This change allows using external secrets which may be created e.g. by [External Secrets](https://external-secrets.io/) tool. I've tried to match convention which is used in [bitnami](https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml#L32) charts.


thjaeckle (approved):
Thanks @kyberpunk for this contribution.
Looks good to me 👍

Issue #457, opened 2023-03-26
Only expose insecure ports in services if explicitly enabled


sophokles73:
The k8s Services created for Hono's components currently always expose insecure ports, regardless of whether the insecure port has been enabled for the component.
The Dispatch Router of the example AMQP 1.0 based messaging infrastructure should support disabling the insecure external port by means of a configuration property.

calohmn:
@sophokles73 Can this be closed?

sophokles73:
Fixed via 506bdc6bfe8170e775e6d8f8b7aaec6488dff67e

PR #458, opened 2023-03-26
Support disabling insecure ports

sophokles73:
This addresses #457

Do not expose insecure service port if insecure port is disabled for a component. Support disabling the insecure port of the Dispatch Router of the AMQP 1.0 based example messaging infrastructure.

### DIFF
--- a/charts/hono/config/router/qdrouterd.json
+++ b/charts/hono/config/router/qdrouterd.json
@@ -36,6 +36,7 @@
     "saslPlugin": "Hono Auth"
   }],
 
+{{- if .Values.amqpMessagingNetworkExample.insecurePortEnabled }}
   ["listener", {
     "host": "0.0.0.0",
     "port": 5672,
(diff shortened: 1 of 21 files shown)


sophokles73:

charts/hono/config/router/qdrouterd.json
```diff
@@ -36,6 +36,7 @@
     "saslPlugin": "Hono Auth"
   }],
 
+{{- if .Values.amqpMessagingNetworkExample.insecurePortEnabled }}
```
calohmn, line 39: This condition can also be placed around the definition of the `amqp` port 5672 in `dispatch-router-deployment.yaml`.

charts/hono/values.yaml
```diff
@@ -350,7 +350,7 @@ adapters:
       # exposed AMQP endpoints.
       # If not set, the adapter by default exposes the secure and insecure ports
       # using an example key and certificate.
-      amqp:
+      amqp: {}
```
calohmn, line 353: For consistency, this change could also be made to the `authServer.hono.auth.amqp` field below, adapting the `hono-service-auth-secret.yaml` accordingly.

calohmn (approved):
LGTM

Issue #466, opened 2023-04-07
Missing connections API endpoint into nginx-configmap.yaml


neottil:
I've install ditto helm charts following this guide https://github.com/eclipse/packages/tree/master/charts/ditto.
When I call the connections API (/api/2/connections) with default basic auth devops:foobar I receive 401 response.

```
curl -v -u devops:foobar http://localhost:8080/api/2/connections
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
* Server auth using Basic with user 'devops'
> GET /api/2/connections HTTP/1.1
> Host: localhost:8080
> Authorization: Basic ZGV2b3BzOmZvb2Jhcg==
> User-Agent: curl/7.87.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 401 Unauthorized
< Server: nginx/1.23.3
< Date: Fri, 07 Apr 2023 14:53:35 GMT
< Content-Type: text/html; charset=utf-8
< Content-Length: 179
< Connection: keep-alive
* Authentication problem. Ignoring this.
< WWW-Authenticate: Basic realm="Authentication required"
< Access-Control-Allow-Methods: GET, POST, PUT, PATCH, DELETE, OPTIONS
< Access-Control-Allow-Credentials: true
< Access-Control-Expose-Headers: *
<
<html>
<head><title>401 Authorization Required</title></head>
<body>
<center><h1>401 Authorization Required</h1></center>
<hr><center>nginx/1.23.3</center>
</body>
</html>
* Connection #0 to host localhost left intact
```

Looking the nginx-configmap.yaml file, the location /api/2/connections is missing. In the nginx.conf file under eclipse-ditto/ditto/deployment/kubernetes/deploymentFiles/nginx (https://github.com/eclipse-ditto/ditto/blob/master/deployment/kubernetes/deploymentFiles/nginx/nginx.conf) this configuration is present in line 31.

thjaeckle:
Indeed that seems be missing from the chart.
We would appreciate a PR, as you already figured out what is missing and how to fix it.

neottil:
I would be delighted to make this contribution. I tried to make a fix, but unfortunately I don't know nginx well and I'm not sure if this is the only problem.

neottil:
Solved by pull request

PR #467, opened 2023-04-11
#466 add connections API endpoint into nginx-configmap.yaml

neottil:
Resolving issue #466.

### DIFF
--- a/charts/ditto/Chart.yaml
+++ b/charts/ditto/Chart.yaml
@@ -17,7 +17,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.2.1
+version: 3.3.0
 appVersion: 3.2.0
 keywords:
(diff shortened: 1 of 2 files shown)


charts/ditto/Chart.yaml
```diff
@@ -17,7 +17,7 @@ description: |
   A digital twin is a virtual, cloud based, representation of his real world counterpart
   (real world “Things”, e.g. devices like sensors, smart heating, connected cars, smart grids, EV charging stations etc).
 type: application
-version: 3.2.1
+version: 3.3.0
```
thjaeckle, line 20: Please adjust the "version" to 3.2.2 - we try to keep at least major+minor version in sync with the appVersion.

And while you are at it: could you please also adjust the `appVersion to 3.2.1`? 
We did a Ditto release 3.2.1 but did not yet adjust the chart.

neottil: Sorry, I edited the minor thinking adding an endpoint was a feature. I immediately correct the version and also the appVersion as indicated

thjaeckle (approved):
LGTM, thx for the contribution

PR #469, opened 2023-05-15
Allow overriding release name

### LINKED ISSUE #430: Allow overriding the hono release name
When deploying hono, it always prefixes the pods (amongst other things) with `.Release.Name`. Almost all other helm charts we use (including Ditto) allow overriding the release name with a custom value.

In our case, ArgoCD sets the Release name to something silly, so we'd like the option to specify the release name ourselves.   
See: https://github.com/eclipse/packages/blob/master/charts/ditto/templates/_helpers.tpl#L14 and https://github.com/eclipse/packages/blob/master/charts/ditto/templates/connectivity-deployment.yaml#L17

WatcherWhale:
This is a continuation of #416 and fixes #430

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 2.4.4
+version: 2.5.0
 # Version of Hono being deployed by the chart
 appVersion: 2.4.0
 keywords:
--- a/charts/hono/README.md
+++ b/charts/hono/README.md
@@ -104,7 +104,7 @@ The command removes all the Kubernetes components associated with the chart and 
 ## Release Notes
 ### 2.5.0
 
-* Allow customizing the pod/service names irrespective or .Release.Name
+* Allow customizing the pod/service names irrespective of .Release.Name.
 
 ### 2.4.4
 
--- a/charts/hono/README.md
+++ b/charts/hono/README.md
@@ -102,6 +102,9 @@ helm uninstall eclipse-hono -n hono
 The command removes all the Kubernetes components associated with the chart and deletes the release.
 
 ## Release Notes
+### 2.5.0
+
+* Allow customizing the pod/service names irrespective or .Release.Name
 
 ### 2.4.4
 
@@ -142,10 +145,6 @@ The command removes all the Kubernetes components associated with the chart and 
 * Update to Kafka chart version 20.x (using Kafka 3.3.x).
 * Update to MongoDB chart version 13.x (using Mongo DB 6.x).
 * Limit Kafka and Zookeeper JVM's memory consumption.
-
-### 2.2.1
-
-* Allow customizing the pod/service names irrespective or .Release.Name
 
 ### 2.2.0
 
--- a/charts/hono/README.md
+++ b/charts/hono/README.md
@@ -143,6 +143,10 @@ The command removes all the Kubernetes components associated with the chart and 
 * Update to MongoDB chart version 13.x (using Mongo DB 6.x).
 * Limit Kafka and Zookeeper JVM's memory consumption.
 
+### 2.2.1
+
+* Allow customizing the pod/service names irrespective or .Release.Name
+
 ### 2.2.0
 
 * Use Hono 2.2.0 container images.
@@ -403,7 +407,7 @@ amqpMessagingNetworkExample:
 
 # mount (existing) Kubernetes secret which contains
 # credentials for connecting to AMQP network
-# into Command Router and protocol adapter containers 
+# into Command Router and protocol adapter containers
 commandRouterService:
   extraVolumes:
   - name: "amqp-network"
--- a/charts/hono/example/add_example_data_device_registry.sh
+++ b/charts/hono/example/add_example_data_device_registry.sh
@@ -11,7 +11,7 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #*******************************************************************************
-HTTP_BASE_URL="http://{{ .Release.Name }}-service-device-registry:8080/v1"
+HTTP_BASE_URL="http://{{ include "hono.fullname" . }}-service-device-registry:8080/v1"
 
 check_status() {
   EXIT_STATUS=$1
--- a/charts/hono/templates/NOTES.txt
+++ b/charts/hono/templates/NOTES.txt
@@ -14,41 +14,41 @@ The output should look similar to this:
 
 NAME                                       {{ "READY   STATUS    RESTARTS   AGE" | indent ( len .Release.Name ) }}
 {{- if .Values.adapters.amqp.enabled }}
-{{ .Release.Name }}-adapter-amqp-65cfb4d675-g5wn4             1/1     Running   0          5m50s
+{{ include "hono.fullname" . }}-adapter-amqp-65cfb4d675-g5wn4             1/1     Running   0          5m50s
 {{- end }}
 {{- if .Values.adapters.coap.enabled }}
-{{ .Release.Name }}-adapter-coap-786b784f47-qx9tg             1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-adapter-coap-786b784f47-qx9tg             1/1     Running   0          5m51s
 {{- end }}
 {{- if .Values.adapters.http.enabled }}
-{{ .Release.Name }}-adapter-http-66bd6bb89c-mng5t             1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-adapter-http-66bd6bb89c-mng5t             1/1     Running   0          5m51s
 {{- end }}
 {{- if .Values.adapters.lora.enabled }}
-{{ .Release.Name }}-adapter-lora-341fca588c-1hl8n             1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-adapter-lora-341fca588c-1hl8n             1/1     Running   0          5m51s
 {{- end }}
 {{- if .Values.adapters.mqtt.enabled }}
-{{ .Release.Name }}-adapter-mqtt-765fcd578b-5rl7n             1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-adapter-mqtt-765fcd578b-5rl7n             1/1     Running   0          5m51s
 {{- end }}
 {{- if ( has "amqp" .Values.messagingNetworkTypes ) }}
-{{ .Release.Name }}-artemis-f8f7dc7f4-864cj                   1/1     Running   0          5m51s
-{{ .Release.Name }}-dispatch-router-6c77dc78bd-hjn4l          1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-artemis-f8f7dc7f4-864cj                   1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-dispatch-router-6c77dc78bd-hjn4l          1/1     Running   0          5m51s
 {{- end }}
 {{- if ( has "kafka" .Values.messagingNetworkTypes ) }}
-{{ .Release.Name }}-kafka-0                                   1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-kafka-0                                   1/1     Running   0          5m51s
 {{- end }}
 {{- if all .Values.deviceRegistryExample.enabled ( eq .Values.deviceRegistryExample.type "mongodb" ) }}
-{{ .Release.Name }}-monogodb-server-77ffb85c97-78554          1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-monogodb-server-77ffb85c97-78554          1/1     Running   0          5m51s
 {{- end }}
-{{ .Release.Name }}-service-auth-84d9695cfc-5wlfh             1/1     Running   0          5m51s
-{{ .Release.Name }}-service-command-router-687fd4fbcf-7527b   1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-service-auth-84d9695cfc-5wlfh             1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-service-command-router-687fd4fbcf-7527b   1/1     Running   0          5m51s
 {{- if .Values.deviceRegistryExample.enabled }}
 {{- if eq .Values.deviceRegistryExample.type "embedded" }}
-{{ .Release.Name }}-service-device-registry-0                 1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-service-device-registry-0                 1/1     Running   0          5m51s
 {{- else }}
-{{ .Release.Name }}-service-device-registry-64d645dd9c-q9c5q  1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-service-device-registry-64d645dd9c-q9c5q  1/1     Running   0          5m51s
 {{- end }}
 {{- end }}
 {{- if ( has "kafka" .Values.messagingNetworkTypes ) }}
-{{ .Release.Name }}-zookeeper-0                               1/1     Running   0          5m51s
+{{ include "hono.fullname" . }}-zookeeper-0                               1/1     Running   0          5m51s
 {{- end }}
 
 Once all pods have reached the READY state, you can start using Hono.
--- a/charts/hono/templates/_helpers.tpl
+++ b/charts/hono/templates/_helpers.tpl
@@ -240,7 +240,7 @@ messaging:
 {{- if .dot.Values.amqpMessagingNetworkExample.enabled }}
   name: {{ printf "Hono %s" .component | quote }}
   amqpHostname: "hono-internal"
-  host: {{ printf "%s-dispatch-router" ( include "hono.fullname" .dot ) | quote }}
+  host: {{ printf "%s-dispatch-router" ( include "hono.fullname" . ) | quote }}
   port: 5673
   keyPath: {{ .dot.Values.adapters.amqpMessagingNetworkSpec.keyPath | quote }}
   certPath: {{ .dot.Values.adapters.amqpMessagingNetworkSpec.certPath | quote }}
--- a/charts/hono/templates/_helpers.tpl
+++ b/charts/hono/templates/_helpers.tpl
@@ -14,6 +14,9 @@
 
 {{/*
 Expand the name of the chart.
+
+The scope passed in is expected to be a dict with keys
+- (mandatory) "dot": the root (".") scope
 */}}
 {{- define "hono.name" -}}
   {{- $nameOverride := .dot.Values.nameOverride -}}
@@ -24,6 +27,9 @@ Expand the name of the chart.
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 If release name contains chart name it will be used as a full name.
+
+The scope passed in is expected to be a dict with keys
+- (mandatory) "dot": the root (".") scope
 */}}
 {{- define "hono.fullname" -}}
   {{- $fullnameOverride := .dot.Values.fullnameOverride -}}
@@ -42,6 +48,9 @@ If release name contains chart name it will be used as a full name.
 
 {{/*
 Create chart name and version as used by the chart label.
+
+The scope passed in is expected to be a dict with keys
+- (mandatory) "dot": the root (".") scope
 */}}
 {{- define "hono.chart" }}
   {{- printf "%s-%s" .dot.Chart.Name .dot.Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
@@ -87,6 +96,9 @@ The scope passed in is expected to be a dict with keys
 
 {{/*
 Add standard labels for resources as recommended by Helm best practices.
+
+The scope passed in is expected to be a dict with keys
+- (mandatory) "dot": the root (".") scope
 */}}
 {{- define "hono.std.labels" -}}
 app.kubernetes.io/name: {{ include "hono.name" . | quote }}
--- a/charts/hono/templates/_helpers.tpl
+++ b/charts/hono/templates/_helpers.tpl
@@ -29,9 +29,6 @@ The scope passed in is expected to be a dict with keys
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 If release name contains chart name it will be used as a full name.
-
-The scope passed in is expected to be a dict with keys
-- (mandatory) "dot": the root (".") scope
 */}}
 {{- define "hono.fullname" -}}
   {{- $fullnameOverride := .Values.fullnameOverride -}}
--- a/charts/hono/templates/_helpers.tpl
+++ b/charts/hono/templates/_helpers.tpl
@@ -16,8 +16,8 @@
 Expand the name of the chart.
 */}}
 {{- define "hono.name" -}}
-  {{- $nameOverride := .Values.global.hono.nameOverride -}}
-  {{- empty $nameOverride | ternary .Chart.Name (tpl $nameOverride $) | trunc 63 | trimSuffix "-" -}}
+  {{- $nameOverride := .dot.Values.nameOverride -}}
+  {{- empty $nameOverride | ternary .dot.Chart.Name (tpl $nameOverride .dot ) | trunc 63 | trimSuffix "-" -}}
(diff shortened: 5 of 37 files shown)


charts/hono/templates/_helpers.tpl
```diff
@@ -16,7 +16,8 @@
 Expand the name of the chart.
```
calohmn, line 16: Would be good to add this here:
```
The scope passed in is expected to be a dict with keys
- (mandatory) "dot": the root (".") scope
```

Also applies to the documentation sections for:
- `define "hono.fullname"`
- `define "hono.chart"`
- `define "hono.std.labels`

charts/hono/templates/_helpers.tpl
```diff
@@ -461,7 +464,7 @@ quarkus:
       exporter:
         otlp:
           {{- if .dot.Values.jaegerBackendExample.enabled }}
-          endpoint: {{ printf "http://%s-jaeger-collector:4317" .dot.Release.Name | quote }}
+          endpoint: {{ printf "http://%s-jaeger-collector:4317" ( include "hono.fullname" .dot ) | quote }}
```
calohmn, line 467: Installing the chart with `--set jaegerBackendExample.enabled=true` causes an error here:
```
Error: INSTALLATION FAILED: template: 
hono/templates/hono-service-device-registry-embedded/hono-service-device-registry-secret.yaml:78:8: 
executing "hono/templates/hono-service-device-registry-embedded/hono-service-device-registry-secret.yaml" 
at <include "hono.quarkusConfig" $args>: 
error calling include: template: hono/templates/_helpers.tpl:467:66: 
executing "hono.quarkusConfig" at <include "hono.fullname" .dot>: 
error calling include: template: hono/templates/_helpers.tpl:29:31: 
executing "hono.fullname" at <.dot.Values.fullnameOverride>: 
nil pointer evaluating interface {}.Values
```

To fix this, it should be 
`include "hono.fullname" .`
above instead of 
`include "hono.fullname" .dot`.

There are 3 more places in `_helpers.tpl` where `include "hono.fullname" .dot` needs to be replaced in the same way.

calohmn:
@WatcherWhale There were changes committed in master (expired Hono Certs getting recreated). Can you rebase?

charts/hono/templates/_helpers.tpl
WatcherWhale: done

charts/hono/templates/_helpers.tpl
WatcherWhale: done

charts/hono/templates/_helpers.tpl
```diff
@@ -263,7 +278,7 @@ The scope passed in is expected to be a dict with keys
 kafka:
 {{- if .dot.Values.kafkaMessagingClusterExample.enabled }}
   commonClientConfig:
-    {{- $bootstrapServers := printf "%[1]s-%[2]s-0.%[1]s-%[2]s-headless:%d" .dot.Release.Name .dot.Values.kafka.nameOverride ( .dot.Values.kafka.service.ports.client | int ) }}
+    {{- $bootstrapServers := printf "%[1]s-%[2]s-0.%[1]s-%[2]s-headless:%d" ( include "hono.fullname" . ) .dot.Values.kafka.nameOverride ( .dot.Values.kafka.service.ports.client | int ) }}
```
calohmn, line 281: In the `bootstrapServers` value, the `fullname` of the Kafka chart deployment has to be used.
This can be achieved by using this:
```
    {{- $kafkaNameValues := pick .dot.Values.kafka "nameOverride" "fullnameOverride" }}
    {{- $kafkaChartDotScope := dict "dot" (dict "Release" .dot.Release "Chart" (dict "Name" "kafka") "Values" $kafkaNameValues) }}
    {{- $bootstrapServers := printf "%[1]s-0.%[1]s-headless:%d" ( include "hono.fullname" $kafkaChartDotScope ) ( .dot.Values.kafka.service.ports.client | int ) }}
```
(Using `deepcopy` on the `.dot` dict caused the linting/deployment to get locked up in my tests.)

charts/hono/templates/_helpers.tpl
```diff
@@ -482,10 +497,10 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- .Values.prometheus.server.fullnameOverride | trunc 63 | trimSuffix "-" -}}
 {{- else -}}
 {{- $name := default "prometheus" .Values.prometheus.nameOverride -}}
-{{- if contains $name .Release.Name -}}
-{{- printf "%s-%s" .Release.Name .Values.prometheus.server.name | trunc 63 | trimSuffix "-" -}}
+{{- if contains $name ( include "hono.fullname" . ) -}}
+{{- printf "%s-%s" ( include "hono.fullname" . ) .Values.prometheus.server.name | trunc 63 | trimSuffix "-" -}}
 {{- else -}}
-{{- printf "%s-%s-%s" .Release.Name $name .Values.prometheus.server.name | trunc 63 | trimSuffix "-" -}}
+{{- printf "%s-%s-%s" ( include "hono.fullname" . ) $name .Values.prometheus.server.name | trunc 63 | trimSuffix "-" -}}
```
calohmn, line 503: The Prometheus service resource is using a name created in the context of the Prometheus chart.
Therefore, using the Hono fullname isn't correct here.
The changes in these lines should just be reverted, using `.Release.Name` again instead of `( include "hono.fullname" . )`.

packages/cloud2edge/Chart.yaml
```diff
@@ -11,7 +11,7 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.4.0
+version: 0.4.1
 appVersion: 0.4.0
```
calohmn, line 15: `appVersion` should be adapted as well.

charts/hono/templates/prometheus/prometheus-config.yaml
```diff
@@ -14,7 +14,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  {{- $args := dict "dot" . "component" "metrics" "name" "prometheus-config" }}
+  {{- $args := dict "dot" (. | deepCopy | merge (dict "Values" (pick .Values.prometheus.server "nameOverride" "fullnameOverride"))) "component" "metrics" "name" "prometheus-config" }}
```
calohmn, line 17: The Prometheus chart expects the ConfigMap name to be `{ReleaseName}-{.Values.prometheus.server.configMapOverrideName}`, i.e. `{ReleaseName}-prometheus-config`.
But the code above doesn't produce that with the current values.yaml.

I would suggest using the following:
```
  {{- /* prometheus chart expects metadata.name to be "{ReleaseName}-{prometheus.server.configMapOverrideName}" */}}
  {{- $prometheusDotScope := . | deepCopy | merge (dict "Values" (dict "fullnameOverride" .Release.Name)) }}
  {{- $args := dict "dot" $prometheusDotScope "component" "metrics" "name" .Values.prometheus.server.configMapOverrideName }}
```

charts/hono/values.yaml
```diff
@@ -1708,7 +1711,7 @@ kafka:
       type: "pem"
       pemChainIncluded: true
       existingSecrets:
-      - "{{ .Release.Name }}-kafka-example-keys"
+      - "{{ include \"hono.fullname\" (dict \"dot\" .) }}-kafka-example-keys"
```
calohmn, line 1714: As the above value is evaluated in the context of the Kafka chart, the `hono.fullname` template will produce `{ReleaseName}-kafka` here, meaning the complete value will be `{ReleaseName}-kafka-kafka-example-keys`.
Therefore, the additional `-kafka' part can be omitted above.

We could also add a comment there, to document the special logic:
```
      - "{{/* evaluated in kafka chart via tpl; kafka chart full name used here! */}}\
      {{ include \"hono.fullname\" (dict \"dot\" .) }}-example-keys"
```
(see also corresponding comment regarding `kafka-example-keys.yaml`)

charts/hono/templates/kafka/kafka-example-keys.yaml
```diff
@@ -14,7 +14,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  {{- $args := dict "dot" . "component" "kafka" "name" "kafka-example-keys" }}
+  {{- $args := dict "dot" (. | deepCopy | merge (dict "Values" (pick .Values.kafka "nameOverride" "fullnameOverride"))) "component" "kafka" "name" "kafka-example-keys" }}
```
calohmn, line 17: Corresponding to the other comment about omitting the duplicate `-kafka` part, this could be changed to:
```
  {{- /* metadata.name is set to "{fullNameOfKafkaSubChart}-example-keys" here, meaning e.g. "{ReleaseName}-kafka-example-keys", not "{ReleaseName}-hono-example-keys" */}}
  {{- $args := dict "dot" (. | deepCopy | merge (dict "Values" (pick .Values.kafka "nameOverride" "fullnameOverride"))) "component" "kafka" "name" "example-keys" }}
```

charts/hono/README.md
```diff
@@ -102,6 +102,10 @@ helm uninstall eclipse-hono -n hono
 The command removes all the Kubernetes components associated with the chart and deletes the release.
 
 ## Release Notes
+### 2.5.0
+
+* Allow customizing the pod/service names irrespective or .Release.Name
```
calohmn, line 107: "irrespective of .Release.Name."

packages/cloud2edge/Chart.yaml
```diff
@@ -11,8 +11,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.4.1
-appVersion: 0.4.1
+version: 0.4.2
+appVersion: 0.4.2
```
calohmn, line 15: @WatcherWhale Could you revert the changes in the cloud2edge Chart.yaml? They are not needed.
(Changes to the Hono version requirement have already been included in PR #483.)

charts/hono/templates/dispatch-router/dispatch-router-secret.yaml
```diff
@@ -20,6 +20,7 @@ metadata:
 type: Opaque
 stringData:
   "qdrouterd.json": |
+    {{- $_ := set $ "dot" $args.dot }}
```
sophokles73, line 23: is there a reason why `$` is not being used anywhere?

charts/hono/templates/hono-service-command-router/hono-service-command-router-deployment.yaml
```diff
@@ -72,4 +72,4 @@ spec:
         {{- include "hono.component.healthChecks" $args | nindent 8 }}
       volumes:
       {{- include "hono.pod.volumes" $args | indent 6 }}
-      serviceAccountName: {{ printf "%s-%s" .Release.Name $args.name | quote }}
+      serviceAccountName: {{ printf "%s-%s" ( include "hono.fullname" . ) $args.name | quote }}
```
sophokles73, line 75: don't we need to pass in `$args` here instead of `.`

charts/hono/example/add_example_data_device_registry.sh
```diff
@@ -11,7 +11,7 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #*******************************************************************************
-HTTP_BASE_URL="http://{{ .Release.Name }}-service-device-registry:8080/v1"
+HTTP_BASE_URL="http://{{ include "hono.fullname" . }}-service-device-registry:8080/v1"
```
sophokles73, line 14: does this actually produce the desired outcome? Don't we need to use a `printf` statement here as well?

charts/hono/values.yaml
```diff
@@ -1708,7 +1711,8 @@ kafka:
       type: "pem"
       pemChainIncluded: true
       existingSecrets:
-      - "{{ .Release.Name }}-kafka-example-keys"
+      # Evaluated in kafka chart via tpl; kafka chart full name used here!
+      - "{{ include \"hono.fullname\" (dict \"dot\" .) }}-example-keys"
```
sophokles73, line 1715: if invoking the template works here, couldn't we also use something like this in order to prevent having to escape the quotes?
```yaml
- {{- printf "%s-example-keys" ( include "hono.fullname" (dict "dot" . ) ) | quote }}
```

sophokles73 (commented):
just some smaller issues
In general, can you also please adapt the copyright headers of the files you changed to reflect the current year?

Thanks for contributing :+1:

charts/hono/templates/_helpers.tpl
```diff
@@ -14,34 +14,46 @@
...
-    {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+  {{- $fullnameOverride := .dot.Values.fullnameOverride -}}
+  {{- if $fullnameOverride  -}}
+    {{- $fullnameOverride | trunc 63 | trimSuffix "-" -}}
   {{- else -}}
-    {{- $name := default .Chart.Name .Values.nameOverride -}}
-    {{- if contains $name .Release.Name -}}
-      {{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+    {{- $nameOverride := .dot.Values.nameOverride -}}
+    {{- $name := empty $nameOverride | ternary .dot.Chart.Name $nameOverride -}}
```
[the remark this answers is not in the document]
WatcherWhale: Not exactly, as `hono.name` is truncated to 63 chars and here we still need the full name to check if the release name matches

calohmn:
@WatcherWhale There has been [an error](https://github.com/eclipse/packages/actions/runs/5974982602/job/16211198131?pr=469) running the checks for this PR:
```
 >>> helm lint charts/hono --values charts/hono/ci/mongodb-based-device-registry-values.yaml
==> Linting charts/hono
Error:  templates/: template: hono/charts/mongodb/templates/standalone/svc.yaml:12:14: 
 executing "hono/charts/mongodb/templates/standalone/svc.yaml" 
at <include "common.labels.standard" (dict "customLabels" .Values.commonLabels "context" $)>: 
error calling include: template: hono/charts/kafka/charts/common/templates/_labels.tpl:6:27: 
 executing "common.labels.standard" at <include "common.names.name" .>: 
error calling include: template: hono/charts/kafka/charts/common/templates/_names.tpl:6:18: 
 executing "common.names.name" at <.Chart.Name>: nil pointer evaluating interface {}.Name
```

This error can be reproduced locally like this:
```
cd charts/hono
helm dependency update
helm lint . --values ci/mongodb-based-device-registry-values.yaml
```
The dependency update will update the used MongoDB chart from `13.16.4` to `13.17.2`.  The new MongoDB chart version makes use of an updated template function in the bitnami `common` dependency version `2.9.0`.
But the referenced Kafka chart is still using an older `common` version which isn't compatible here, and that older `common` chart version seems indeed to be loaded. Hence the error above.

See https://github.com/bitnami/charts/issues/18890#issuecomment-1694733376.

The straightforward solution seems to be to prevent the usage of the MongoDB `13.17.x` chart version by using
```
- name: "mongodb"
  repository: "https://charts.bitnami.com/bitnami"
  version: "~13.16.x"
```
in the Hono `Chart.yaml`.

The alternative would be to update the Kafka chart to `25.0.0` (released last week). This version is also using `common` `2.9.0`.
But I think we should leave this Kafka chart version jump to a separate PR.

charts/hono/templates/hono-service-command-router/hono-service-command-router-secret.yaml
calohmn, line 1: The line 
`{{- tpl ( .Files.Get "config/command-router/embedded-cache-config.xml" ) . | nindent 4 }}` 
in this file needs to be replaced with
```
  {{- $_ := set $ "dot" $args.dot }}
  {{- tpl ( .Files.Get "config/command-router/embedded-cache-config.xml" ) $ | nindent 4 }}
```
.

charts/hono/templates/NOTES.txt
```diff
@@ -14,41 +14,41 @@ The output should look similar to this:
 
 NAME                                       {{ "READY   STATUS    RESTARTS   AGE" | indent ( len .Release.Name ) }}
 {{- if .Values.adapters.amqp.enabled }}
-{{ .Release.Name }}-adapter-amqp-65cfb4d675-g5wn4             1/1     Running   0          5m50s
+{{ include "hono.fullname" . }}-adapter-amqp-65cfb4d675-g5wn4             1/1     Running   0          5m50s
```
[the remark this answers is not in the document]
calohmn: Actually, I think we could get rid of this whole rather obscure `{{- $_ := set $ "dot" $args.dot }}` usage, by reverting the `define "hono.fullname"` named template changes - not requiring a `dot` dict there anymore but letting "hono.fullname" use `.Values` directly, instead of `.dot.Values`.

That would mean:
- replacing `include "hono.fullname" $args` with `include "hono.fullname" .` in the yaml files
- reverting the above `$_ := set $ "dot"` changes
- replacing `include "hono.fullname" .` with `include "hono.fullname" .dot` in `_helpers.tpl`
- adapting the place where `$kafkaChartDotScope` is defined within `_helpers.tpl`, removing the  outer `dict "dot"`
- replacing `include "hono.fullname" (dict "dot" .)` with `include "hono.fullname" .` in `values.yaml`
- adapting `kafka-example-keys.yaml`, using this in the `metadata` block:
```
  {{/* metadata.name is set to "{fullNameOfKafkaSubChart}-example-keys" here, meaning e.g. "{ReleaseName}-kafka-example-keys", not "{ReleaseName}-hono-example-keys" */}}
  {{- $kafkaNameValues := pick .Values.kafka "nameOverride" "fullnameOverride" }}
  {{- $kafkaChartDotScope := dict "Release" .Release "Chart" (dict "Name" "kafka") "Values" $kafkaNameValues }}
  {{- $args := dict "dot" ($kafkaChartDotScope) "component" "kafka" "name" "example-keys" }}
```

@WatcherWhale @sophokles73 WDYT?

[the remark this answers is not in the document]
WatcherWhale: It is indeed a rather obscure line, I'm in favor of changing it.

charts/hono/Chart.yaml
```diff
@@ -46,7 +46,7 @@ dependencies:
   condition: "grafana.enabled"
 - name: "mongodb"
   repository: "https://charts.bitnami.com/bitnami"
-  version: "^13.x"
+  version: "~13.16.x"
```
sophokles73, line 49: is this related to the issue of overriding the release name?
If not, I would prefer to do any necessary changes to the Mongo DB version in a separate issue/PR ...

WatcherWhale: It is to fix this issue https://github.com/eclipse/packages/pull/469#issuecomment-1696238963

calohmn: Created separate PR #491 for that.

charts/hono/templates/NOTES.txt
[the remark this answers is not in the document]
calohmn: The files
`charts/hono/templates/example-data-grid/configmap.yaml`
and
`charts/hono/templates/hono-service-device-registry-base/hono-service-device-registry-post-install-configmap.yaml`
still need to be adapted (`{{- $_ := set $ "dot" $args.dot }}` still in there).

charts/hono/templates/_helpers.tpl
```diff
@@ -14,21 +14,32 @@
...
+  {{- empty $nameOverride | ternary .dot.Chart.Name $nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 If release name contains chart name it will be used as a full name.
+
+The scope passed in is expected to be a dict with keys
+- (mandatory) "dot": the root (".") scope
```
calohmn, line 34: This should be removed again.

calohmn:
@WatcherWhale Could you rebase and resolve the conflicts? PR #491 has been merged.

charts/hono/templates/_helpers.tpl
```diff
@@ -14,9 +14,15 @@
 
 {{/*
 Expand the name of the chart.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+
+The scope passed in is expected to be a dict with keys
+- (mandatory) "dot": the root (".") scope
 */}}
 {{- define "hono.name" -}}
```
calohmn, line 22: I think it would be consistent to then also remove the "dot" dict usage in the "hono.name" and "hono.chart" named templates again.

charts/hono/templates/NOTES.txt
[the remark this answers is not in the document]
calohmn: In these 2 files, can you also restore usage of `.` instead of `$` in the `{{- tpl ( .Files.Get ...` line?

WatcherWhale:
@sophokles73 are there any other changes, I still need to apply for the 2 remaining unresolved issues?

charts/hono/templates/_helpers.tpl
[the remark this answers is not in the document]
sophokles73: I see

sophokles73:
> are there any other changes, I still need to apply for the 2 remaining unresolved issues?

Well, as my comment, indicates, IMHO it would be good if you added a comment explaining what is actually being computed, as it is not really obvious to me ...

charts/hono/templates/_helpers.tpl
```diff
@@ -14,34 +14,46 @@
...
 {{/*
 Expand the name of the chart.
+
+The scope passed in is expected to be a dict with keys
+- (mandatory) "dot": the root (".") scope
 */}}
 {{- define "hono.name" -}}
-  {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+  {{- $nameOverride := .dot.Values.nameOverride -}}
+  {{- empty $nameOverride | ternary .dot.Chart.Name $nameOverride | trunc 63 | trimSuffix "-" -}}
```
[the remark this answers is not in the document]
WatcherWhale: I updated the comment and also restructured the code a bit to make it more legible and similar to what is used in `hono.fullname`

packages/cloud2edge/Chart.yaml
```diff
@@ -11,8 +11,8 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 apiVersion: v1
-version: 0.4.2
-appVersion: 0.4.2
+version: 0.4.1
+appVersion: 0.4.1
```
sophokles73, line 15: I do not think that the c2e package chart needs to be changed as part of this PR, or am I mistaken?

sophokles73 (changes requested):

packages/cloud2edge/Chart.yaml
calohmn: I guess that was a rebase-error - the c2e chart had been updated to 0.4.2 in the meantime.

WatcherWhale: It was indeed a rebase error I fixed it now

charts/hono/Chart.yaml
```diff
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 2.4.4
+version: 2.5.0
```
sophokles73, line 18: one final question:

Based on the information from the [When to increase a chart’s version number? section](https://www.eclipse.org/packages/contribute/) of the IoT Packages contribution guide, I would think that increasing the patch version should be sufficient for this kind of change, or am I missing something?

@calohmn WDYT?

calohmn: The changes in this PR change the name of the Hono resources if the chosen _release name_ doesn't contain "hono" (or the corresponding `nameOverride` for hono).
So, using the "eclipse-hono" release name (as documented in the Hono chart README), there would be no change to most of the names (only the Kafka related resources, I think).

But, when deploying Hono as part of the cloud2edge package, using `c2e` as release name, the names will change from
```
c2e-adapter-amqp
c2e-service-auth
c2e-ditto-gateway
c2e-ditto-nginx
```
to
```
c2e-hono-adapter-amqp
c2e-hono-service-auth
c2e-ditto-gateway
c2e-ditto-nginx
```
So, this is an incompatible change, requiring adaptations to the cloud2edge chart. That's why we have restricted the Hono chart version in the cloud2edge dependencies to `~2.4.0`.

So, we have to (at least) increase the Hono chart minor version here. (See also discussion in https://github.com/eclipse/packages/pull/416#discussion_r1001436531.)

sophokles73 (approved):
LGTM

calohmn (approved):
@WatcherWhale Thanks for bringing this to completion!

PR #471, opened 2023-06-13
Deprecate Ditto Helm chart at iot-packages

thjaeckle:
Hey all.

The Ditto Helm chart is now maintained within the [Ditto GitHub repository](https://github.com/eclipse-ditto/ditto/tree/master/deployment/helm) and distributed via [Docker Hub's Ditto repository](https://hub.docker.com/r/eclipse/ditto).

In order to avoid duplicated effort in maintaining 2 charts, it is therefore deprecated here.

As part of the inception of the new Ditto chart, a lot of configuration options were added to it in order to make it easier than ever to configure it for productive setups, which the old chart was never very strong in - not being used by any backing company for productive use.

Would be nice to get an approval, if someone is still active on this project (which is another reason to move out).

Best regards
Thomas


ctron (approved):

ctron:
I will merge it anyway!

Issue #482, opened 2023-08-09
[cloud2edge] No connection from Hono device registry to MongoDB


calohmn:
Installing the cloud2edge chart with the default configuration from the `values.yaml` leads to the Hono service registry not being able to connect to the MongoDB. The device registry is using the wrong hostname (`ditto-mongodb`) for connecting to the MongoDB (available under `<release-name>-mongodb`, e.g. `c2e-mongodb`).

Apart from that, the MongoDB pod is also not getting ready (container getting OOM killed and using a lot of CPU).

calohmn:
The first issue seems to be due to [this change](https://github.com/eclipse/packages/pull/439/files#diff-1279b106274eb06e230785de7be0cf045f75be3d4c240de7a22751e90333f781R712) in the ditto chart, letting a non-fixed mongodb hostname to be used - which the Hono device registry isn't using.

The MongoDB pod CPU/memory issues seem to be due to a MongoDB chart update, letting #403 come up again.

PR #483, opened 2023-08-10
[#482] Fix wrong MongoDB hostname; fix MongoDB health checks.

calohmn:
Fix for #482 .
Fixes Hono devices registry not being able to connect to the MongoDB instance because it used the wrong hostname. Also fixes MongoDB not getting ready because the health checks took too long and caused OOM errors.
Also restrict Hono chart dependency version range (needed for upcoming changes addressing #430).


calohmn:
@sophokles73 Could you have a look?

sophokles73 (approved):
LGTM

Maybe we could use the tool suggested in https://github.com/bitnami/charts/issues/10264#issuecomment-1573530231 for doing the healthcheck instead?

However, this is something that we might want to add to the Hono and Ditto charts in separate PRs ...

PR #491, opened 2023-08-29
[#490] Pin MongoDB chart version, apply minor fixes.

### LINKED ISSUE #490: Error installing/linting local Hono chart with latest dependency versions
From https://github.com/eclipse/packages/pull/469#issuecomment-1696238963:
There is currently an error when trying to install the Hono chart from local sources using the latest dependencies. 

The error can already be seen when running `helm lint`:
```
cd charts/hono
helm dependency update
helm lint . --values ci/mongodb-based-device-registry-values.yaml
```
Output:
```
Error:  templates/: template: hono/charts/mongodb/templates/standalone/svc.yaml:12:14: 
 executing "hono/charts/mongodb/templates/standalone/svc.yaml" 
at <include "common.labels.standard" (dict "customLabels" .Values.commonLabels "context" $)>: 
error calling include: template: hono/charts/kafka/charts/common/templates/_labels.tpl:6:27: 
 executing "common.labels.standard" at <include "common.names.name" .>: 
error calling include: template: hono/charts/kafka/charts/common/templates/_names.tpl:6:18: 
 executing "common.names.name" at <.Chart.Name>: nil pointer evaluating interface {}.Name
```

Analysis:
The dependency update will update the used MongoDB chart from `13.16.4` to `13.17.2`.  The new MongoDB chart version makes use of an updated template function in the bitnami `common` dependency version `2.9.0`.
But the referenced Kafka chart is still using an older `common` version which isn't compatible here, and that older `common` chart version seems indeed to be loaded. Hence the error above.

See https://github.com/bitnami/charts/issues/18890#issuecomment-1694733376.

The straightforward solution seems to be to prevent the usage of the MongoDB `13.17.x` chart version by using
```
- name: "mongodb"
  repository: "https://charts.bitnami.com/bitnami"
  version: "~13.16.x"
```
in the Hono `Chart.yaml`.

The alternative would be to update the Kafka chart to `25.0.0` (released last week). This version is also using `common` `2.9.0`.

calohmn:
Fixes #490.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 2.4.3
+version: 2.4.4
 # Version of Hono being deployed by the chart
 appVersion: 2.4.0
 keywords:
@@ -46,7 +46,7 @@ dependencies:
   condition: "grafana.enabled"
 - name: "mongodb"
   repository: "https://charts.bitnami.com/bitnami"
-  version: "^13.x"
+  version: "~13.16.x"
   condition: "mongodb.createInstance"
 - name: "kafka"
   repository: "https://charts.bitnami.com/bitnami"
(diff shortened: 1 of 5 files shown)


sophokles73 (approved):
LGTM

Issue #486, opened 2023-08-24
Cloud2Edge: Ditto helm chart moved to Ditto's repository


sgloutnikov:
I just noticed that the [Ditto helm chart](https://github.com/eclipse/packages/tree/master/charts/ditto) used in cloud2edge [requirements.yaml](https://github.com/eclipse/packages/blob/master/packages/cloud2edge/requirements.yaml) is no longer maintained in this repository but was moved to [Ditto's repository](https://github.com/eclipse-ditto/ditto/tree/master/deployment/helm/ditto) and is also served from a different Helm repository. Does this mean that the cloud2edge package should be updated to use the new and updated Ditto Helm chart, or are there some incompatibilities that could be preventing the switch over?

I'll test on my end and see how far I can get with updating the cloud2edge Ditto version to latest.

casafurix:
Hello, I am actually unable to install the cloud2edge package using the steps mentioned on the website, can you please guide me the right steps now to install it? I want to use Ditto+Hono. Thanks

calohmn:
(Cloud2edge installation issues were resolved with @casafurix in #481. In general, the [Gitter channel](https://gitter.im/eclipse/packages) or Stackoverflow (with [eclipse-iot](https://stackoverflow.com/questions/tagged/eclipse-iot) tag) are good places for asking questions, or, in case of a chart or documentation bug, an issue can be raised here.)

calohmn:
@sgloutnikov yes, the cloud2edge chart should be updated. A PR is welcome here.

calohmn:
@sgloutnikov The update actually looked quite straightforward. I've integrated it in PR #493 .

sgloutnikov:
Really strange, I'm unable to bring up the environment in a fresh k3s or minikube single node cluster, any tips are greatly appreciated. This behaviour seems identical for k3s with LoadBalancer or minikube without. 

Looks like a bunch of pods failing with rediness probe, which chains down to nginx and swagger-ui not starting up. Lastly the gateway pod rediness probe looks like is missing its host, and only port is being configured. Some details bellow:

Helm install:
```
 cloud2edge git:(pr/493) helm install -n $NS --wait --timeout 20m --set hono.useLoadBalancer=true --set ditto.nginx.service.type=LoadBalancer $RELEASE ./
```

Get pods:
```
 ➜  ~ k get pods
NAME                                           READY   STATUS                  RESTARTS        AGE
[...]

Issue #432, opened 2022-11-11
C2E "Tour" documentation is outdated, still refers to AMQP 1.0 connections


thjaeckle:
The current [Create a new connection from Ditto to Hono](https://www.eclipse.org/packages/packages/cloud2edge/tour/#create-a-new-connection-from-ditto-to-hono) still refers to creating a AMQP 1.0 connection to Hono.

AFAIK the c2e package however moved to Apache Kafka based connectivity from Ditto to Hono.
The tour documentation should be adjusted to reflect that change.

@kalinkostashki could you have a look please?

kalinkostashki:
Yes, I will take a look and try to fix where appropriate :)

ishwar25667:
Hi @thjaeckle ,
I am using c2e package & tried to establish connection with AMQP 1.0 but getting below issue.

{"status":504,"error":"connectivity:connection.failed","message":"Failed to consume sources: [command_response/my-tenant/replies-0, telemetry/my-tenant-0, event/my-tenant-0]","description":"command_response/my-tenant/replies-0: no such node [condition = amqp:not-found], telemetry/my-tenant-0: message address must not contain tenant ID only [condition = hono:bad-request], event/my-tenant-0: message address must not contain tenant ID only [condition = hono:bad-request]"}

{"status":504,"error":"connectivity:connection.failed","message":"Failed to consume sources: [telemetry/my-tenant/my-device-1-0, event/my-tenant/my-device-1-0]","description":"telemetry/my-tenant/my-device-1-0: no such node [condition = amqp:not-found], event/my-tenant/my-device-1-0: no such node [condition = amqp:not-found]"}

{
  "status": 504,
  "error": "connectivity:connection.failed",
  "message": "Failed to consume sources: [telemetry/org.acme/my-device-1/things/twin/commands/modify-0]",
  "description": "telemetry/org.acme/my-device-1/things/twin/commands/modify-0: can only publish for device of same tenant [condition = amqp:unauthorized-access]"
}

Using below documentation-

https://www.eclipse.org/packages/packages/cloud2edge/tour/#create-a-new-connection-from-ditto-to-hono

Can you please let me know how I can fix this issue?

kalinkostashki:
Hi @ishwar25667 ,

I still haven't gotten around to fixing the documentation, but kafka is the default messaging used between Hono and Ditto for a long time. 
Are you using the default deployment of cloud2edge? If so there would be no AMQP messaging there.

calohmn:
[...]

Issue #492, opened 2023-08-31
Kafka connection in cloud2edge


bs3245:
I have installed cloud2edge package in mickrok8s and I am able to create tenants, devices and credentials. 
 
Can you provide me the steps on how to enable kafka based Hono connection in ditto in cloud2edge?

Also can you suggest how to implement inbound and outbound kafka connection?

@calohmn could you please take this question
*This is more of an question than issue, I don't know to how to add question tag*

calohmn:
@bs3245 When deploying the current cloud2edge chart version (0.4.2), there is already a Kafka connection being created for the `org.eclipse.packages.c2e` tenant.
As for creating another connection for a new tenant, I've created PR #493 to update the cloud2edge tour webpage accordingly.
(In the PR code there is currently still an issue concerning the Kafka topic mapping - I'm looking into that.)

There is already #468 concerning issues creating a new tenant and connection, and there is #432 about updating the cloud2edge tour with instructions how to add a Kafka connection.

I'm closing this as a duplicate then, you can follow the PR #493 to see when the cloud2edge documentation gets updated.

In general, if you have a question, you can ask it in the [Gitter channel](https://gitter.im/eclipse/packages).

calohmn:
Duplicate of #468

bs3245:
@calohmn So can you tell me is the any other way for connecting hono to ditto if not amqp(outdated) and kafka(not ready) for new tenant?

calohmn:
@bs3245 Connecting Hono to Ditto can be done via AMQP 1.0 or Kafka. I'm positive we will get the Kafka connection example in the cloud2edge chart fully working soon together with the new Ditto version (see #493).

Concerning the AMQP connection support, I can say with respect to Hono that having started out with AMQP 1.0, we added Kafka support in Hono 2.0. Kafka has been the focus of development since then, mainly because of its wider adoption.

bs3245:
Hi @calohmn,

I'm trying to install the new c2e version, do I need add any helm repo since ditto has been moved to its own repository?

calohmn:
@bs3245 No, that is not needed. You just need the added eclipse-iot repo, as described [here](https://www.eclipse.org/packages/repository/).
To make sure you use the newest cloud2edge chart version, run `helm repo update eclipse-iot`.
To install the newest cloud2edge chart version, you need at least Helm version 3.8.

bs3245:
[...]

PR #493, opened 2023-09-07
[#486] [#432] Update Ditto in cloud2edge chart, document creating Kafka connection in c2e tour

calohmn:
Includes these commits:

- [#432]
Update ditto connection chapter in cloud2Edge tour.
Adding the request for creating a Kafka connection.

- [#486]
Update Ditto chart version in cloud2edge chart.
Using the Ditto chart from the Ditto repository now.
Also updated the tour to add instructions on using
the Ditto UI.

- Update cloud2Edge chart version to 0.5.0.

### DIFF
--- a/homepage/_packages/cloud2edge/tour.md
+++ b/homepage/_packages/cloud2edge/tour.md
@@ -197,7 +197,7 @@ This should return a result of `201 Created`.
     etag: be547a07-4a03-4c43-a274-d02f63f8d467
     location: /v1/tenants/my-tenant
     content-type: application/json; charset=utf-8
-    content-length: 12
+    content-length: 18
     
     {"id":"my-tenant"}
 {% enddetails %}
@@ -215,9 +215,9 @@ This should return a result of `201 Created`.
 {% details Example of a successful result %}
     HTTP/1.1 201 Created
     etag: d48f4e13-b398-4c73-bbc3-5ac97a81b3e8
-    location: /v1/devices/iot/org.acme:my-device-1
+    location: /v1/devices/my-tenant/org.acme:my-device-1
     content-type: application/json; charset=utf-8
-    content-length: 17
+    content-length: 29
     
     {"id":"org.acme:my-device-1"}
 {% enddetails %}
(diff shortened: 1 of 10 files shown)


calohmn:
There is still an issue following all the steps in the tour concerning the manually added connection - the Kafka topic name contains an extra tenant-name prefix where just the device-id should be.
I'm looking into that...

thjaeckle (approved):
Looks good to me. Thanks for updating, @calohmn :+1:

homepage/_packages/cloud2edge/tour.md
```diff
@@ -488,6 +694,20 @@ In order to add more twins, we simply create additional devices via ["Register a
...
+following command can be used:
+
+{% clipboard %}
+curl -i -k -u my-auth-id-1@my-tenant:my-password -H 'application/json' --data-binary '{
+  "topic": "org.acme/my-device-1/things/twin/commands/modify",
+  "headers": {},
+  "path": "/features/temperature/properties/value",
+  "value": 45
+}' ${HTTP_ADAPTER_BASE_URL:?}/telemetry
+{% endclipboard %}
```
calohmn, line 709: Having a closer look in the Jaeger Tracing UI about how this telemetry request is processed, I see an error.

send_telemetry__wrong_connection
In the `kafka_consume` step, there is still the right Ditto connection. 
But in the `connectivity_message_mapping` step, the connection that was created for the cloud2edge default `org.eclipse.packages.c2e` tenant is used - so the wrong connection.
Following that, there is a 404 error trying to find the device in this `org.eclipse.packages.c2e` tenant.

Strangely, when I had first created this twin _without a policy_ (as a workaround for the policy creation error - see my other comment), the same telemetry request didn't show this wrong connection usage.

@thjaeckle Do you have an idea here?

---
To enable tracing, I used these chart parameters:
```
hono:
  jaegerBackendExample:
    enabled: true
ditto:
  global:
    tracing:
      enabled: true
      otelExporterOtlpEndpoint: "http://c2e-jaeger-collector:4317"
      sampler: always
```
The URL of the Jaeger UI is in the `JAEGER_QUERY_BASE_URL` env variable set by the `setCloud2EdgeEnv.sh` script.

bs3245:
@calohmn So can you tell me is the any other way for connecting hono to ditto if not amqp(outdated) and kafka(not ready) for new tenant at this moment?

Can you also say why amqp connection is outdated?

homepage/_packages/cloud2edge/tour.md
thjaeckle: That is expected if both connections use the same authorizationContext.
A connection target will not be notified about events which were the result of a command consumed via the source of the same connection.

However if a second connection exists which has the permissions to read the updated thing, then this connection will emit the event to the configured target

Each connection for different Hono tenants should use a separate authorizationContext string to prevent that.

homepage/_packages/cloud2edge/tour.md
```diff
@@ -409,7 +615,7 @@ curl -i -X PUT -u ditto:ditto -H 'Content-Type: application/json' --data '{
   "entries": {
     "DEFAULT": {
       "subjects": {
-        "{%raw%}{{ request:subjectId }}{%endraw%}": {
```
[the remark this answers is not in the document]
thjaeckle: No, nothing changed here.. Should still work.
Can you reproduce that also with postman?

calohmn:
@bs3245 Let's stick to the existing discussion in #492.

homepage/_packages/cloud2edge/tour.md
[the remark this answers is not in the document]
calohmn: Oh, sorry, I copied the curl request from the text view of the c2e-tour markdown file, not paying attention to the `{%raw%}` formatting instruction in there.

homepage/_packages/cloud2edge/tour.md
calohmn: So, I just replace all occurrences of `pre-authenticated:hono-connection` with `pre-authenticated:hono-connection-my-tenant` in the connection JSON and the JSON of the policy concerning that tenant?

thjaeckle: Yes - if those are 2 separate tenants they should use different "pre authentication" in connections and in their policies.

calohmn: Thanks, I've made that change and it's working now.

PR #495, opened 2023-09-07
[#494] add priorityClass support

### LINKED ISSUE #494: Add priorityClass support
Add support to set a priorityClass in the hono charts. This will allow more control over the scheduling of pods.

WatcherWhale:
Closes #494

### DIFF
--- a/charts/hono/templates/artemis/artemis-deployment.yaml
+++ b/charts/hono/templates/artemis/artemis-deployment.yaml
@@ -30,6 +30,7 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- include "hono.pod.priorityClassName" $args | nindent 6 }}
       containers:
       - name: apache-activemq-artemis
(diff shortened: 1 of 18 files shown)


charts/hono/templates/artemis/artemis-deployment.yaml
```diff
@@ -30,6 +30,7 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      priorityClassName: "{{ $args.componentConfig.pod.priorityClassName }}"
```
calohmn, line 33: I think it would be better to put this in a
`{{- if $args.componentConfig.pod.priorityClassName }}`
block.

And, the quoting is usually done via the `quote` template function in the Hono chart:
`priorityClassName: {{ $args.componentConfig.pod.priorityClassName | quote }}`

calohmn: As this is repeated often, a named template in `_helpers.tpl` could be added.

calohmn (approved):
LGTM

Thanks for contributing!

PR #497, opened 2023-09-08
[#496] Allow affinity to be configured

### LINKED ISSUE #496: Add affinity configuration in hono chart

WatcherWhale:
Closes #496


charts/hono/values.yaml
```diff
@@ -1188,6 +1194,7 @@ deviceRegistryExample:
     pod:
       labels: {}
       annotations: {}
+      affinity: {}
```
calohmn, line 1197: This value still needs to be used in 
`hono-service-device-registry-embedded/hono-service-device-registry-statefulset.yaml`

charts/hono/templates/artemis/artemis-deployment.yaml
```diff
@@ -86,4 +86,5 @@ spec:
         {{- include "hono.container.volumeMounts" $args | indent 8 }}
       volumes:
       {{- include "hono.pod.volumes" $args | indent 6 }}
+      {{- include "hono.pod.affinity" $args | nindent 6 }}
```
calohmn, line 89: Having the `affinity` entry at the bottom of the `spec` items list seems somewhat non-intuitive to me. Could you move it up (somewhere above the (init)containers ?

calohmn (approved):
LGTM

Thanks for contributing!

PR #498, opened 2023-09-09
[#486] Use newer Helm version in Jenkinsfile.

calohmn:
Helm >= 3.8 needed to install cloud2edge chart now.

### DIFF
--- a/.jenkins/Jenkinsfile
+++ b/.jenkins/Jenkinsfile
@@ -8,7 +8,7 @@ properties([
 ])
 
 def label = "packages-${UUID.randomUUID().toString()}"
-def HELM_VERSION = "3.5.4"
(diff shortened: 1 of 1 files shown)


sophokles73:
How do we know that it works with Helm 3.8.x if we use 3.12.x now?

calohmn:
I think we can settle with defining 3.9 (from May 2022) as the minimum Helm version.
I've updated the commit here to use 3.9.2 and the commit in #502 for the `ci.yaml` as well.

calohmn:
The new Ditto chart referred to in #486 is being referenced from an OCI-based registry, and [OCI support in Helm](https://helm.sh/docs/topics/registries/) is only generally available in Helm version 3.8.

sophokles73 (approved):
LGTM

PR #500, opened 2023-09-09
[#499] [#501] Cloud2edge: Fix ditto resource config, use Connections API.

### LINKED ISSUE #499: Cloud2edge chart ditto resource config not matching new ditto chart
The new Ditto chart used in the cloud2edge chart has a different way of configuring the CPU and memory resources compared to the older Ditto chart.

The cloud2dge `values.yaml` still needs to be adapted to that - otherwise the chart has higher resource requirements than documented in the cloud2edge chart installation instructions.

calohmn:
This fixes
- #499
- #501


thjaeckle (approved):
Great work.
Looks good to me 👍

PR #509, opened 2023-09-22
[#512] Add service account for Hono adapter pods; update tracing image versions.

### LINKED ISSUE #512: Set service account on Hono protocol adapter pods
The upcoming Hono 2.5 and 2.4.x, 2.3.x releases will contain logic to determine the K8s container id from the protocol adapter pods via the K8s API if cgroups v2 is used (see https://github.com/eclipse-hono/hono/issues/3537).

For that, the Hono chart needs to be changed, setting a service account on the protocol adapter pods that has a role binding that allows 'get' operations on the 'pods' resources.

calohmn:
This fixes #512.
Apply changes needed for https://github.com/eclipse-hono/hono/pull/3546, adding corresponding service accounts.

Also update jaegertracing/all-in-one and opentelemetry-collector image versions and fix Hono chart icon URL.

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -27,7 +27,7 @@ keywords:
 home: https://www.eclipse.org/hono/
 sources:
 - https://github.com/eclipse-hono/hono
-icon: https://eclipse.dev/hono/img/favicon.ico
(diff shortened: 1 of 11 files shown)


sophokles73 (approved):
LGTM

Issue #510, opened 2023-09-23
Grafana dashboards empty if non-default Hono chart release name used


calohmn:
When installing the Hono chart with a non-default release name (i.e. not "eclipse-hono" as used in the documentation) that doesn't contain the string "hono", the Grafana dashboards are empty.
This is a regression introduced via the changes made for #430.

Issue #507, opened 2023-09-22
Use Bitnami OCI packages in Hono chart dependencies


calohmn:
The standard way in which Bitnami charts are deployed is now via OCI packages (see announcement [here](https://blog.bitnami.com/2023/04/httpsblog.bitnami.com202304bitnami-helm-charts-now-oci.html)). The Hono chart dependencies should be updated accordingly.

PR #511, opened 2023-09-23
[#506][#507][#510] Apply Hono chart fixes/improvements

### LINKED ISSUE #506: Enable using non-Hono-chart mongoDB with name based on ReleaseName
When installing the Hono chart as part of a package that already includes a MongoDB installation (like in the cloud2edge chart, where the Ditto mongoDB is used), the name of the mongoDB host currently has to be set to a fixed value.
E.g.
```
  deviceRegistryExample:
    type: mongodb
    mongoDBBasedDeviceRegistry:
      mongodb:
        host: ditto-mongodb
```
It would be better however, if the name of the mongoDB host could contain the release name for example.
In the case of the cloud2edge chart, this would enable the mongoDB resources to have the same naming scheme as the other resoures.

A solution here would be to let the `deviceRegistryExample.mongoDBBasedDeviceRegistry.mongodb` map be evaluated via the `tpl` function.
The host could then be configured as
`host: '{{ .Release.Name }}-mongodb'`.

calohmn:
This fixes #506, #507, #510 and includes further minor fixes and improvements.

Included commits:

- [#510] Fix pod-selector in prometheus-config.yaml.
- [#506] Apply tpl function on mongoDBBasedDeviceRegistry mongodb values.
  Enabling the configured mongoDB host to contain the release name reference.
- [#507] Use Bitnami OCI packages in Hono chart dependencies.
- Fix wrong links, apply README corrections.
- Rename profile yaml files to *-values.yaml.
This lets these files be detected by the IntelliJ IDEA Kubernetes plugin.
- Simplify subchart Values references, remove superfluous template params.
Using `.Subcharts.<chart-name>` (see https://github.com/helm/helm/pull/9957) instead of building a dict with subchart Values manually.


sophokles73 (approved):
LGTM

PR #513, opened 2023-10-12
Update c2e chart to Hono 2.5, add tracing profile, improve tour

### LINKED ISSUE #505: Update cloud2edge chart to use Hono chart >= 2.5.0
Hono chart version 2.5.0 contains changes to the Hono resource names in case a release name not containing "hono" is used (see #469).

Therefore the cloud2edge chart and the cloud2edge tour need to be adapted.

calohmn:
This fixes #505 .

Includes these commits:
- [#505] Use Hono chart 2.5.5 in cloud2edge chart.
This includes changes to the naming of the Hono resources (now having `{release-name}-hono` as prefix instead of `{release-name}`). Also update Ditto chart to version 3.3.7.
- Rename Ditto MongoDB resources in cloud2edge chart.
Now using `{release-name}-mongodb` as prefix instead of `ditto-mongodb`.
- Add tracing profile in cloud2edge chart.
Also rename other profile yaml files to *-values.yaml. This lets these files be detected by the IntelliJ IDEA Kubernetes plugin.
- Improve getting `NODE_IP` in `setCloud2EdgeEnv.sh`.
First check the ExternalIP of the Kubernetes node.
- Add info about using mosquitto_sub in cloud2edge tour.
Also update `setCloud2EdgeEnv.sh` to set `MOSQUITTO_OPTIONS` and write `TRUSTSTORE_PATH` file.
- Update cloud2Edge chart version to 0.7.0; add release notes.


packages/cloud2edge/Chart.yaml
```diff
@@ -25,9 +25,16 @@ keywords:
 home: https://eclipse.org/packages/packages/cloud2edge/
 icon: https://eclipse.org/packages/packages/cloud2edge/images/c2e-logo.svg
 sources:
-  - https://github.com/eclipse/packages/packages/cloud2edge/
+  - https://github.com/eclipse/packages/tree/master/packages/cloud2edge
```
sophokles73, line 28: FMPOV this should point to the standard repository URL

calohmn: To me the link to the actual chart source directory looks more convenient. Going to the parent repository from there is trivial.
Starting out with the repository URL however, it's not directly obvious in which directory the cloud2edge chart is defined.
(Kafka and other Bitnami charts also link to the chart sub-directory in the sources link, btw.)

sophokles73: Ok, now I see your point. I was under the impression that the original link actually pointed to the correct place but it seems to result in a 404 ...

sophokles73 (approved):
LGTM

PR #514, opened 2023-10-12
Update to kafka 3.4

sophokles73:
Updated to bitnami/kafka chart version 21.x which uses Kafka 3.4.
This aligns better with the Kafka version being used for testing in Hono
2.4.

### DIFF
--- a/charts/hono/example/add_example_data_device_registry.sh
+++ b/charts/hono/example/add_example_data_device_registry.sh
@@ -11,7 +11,9 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #*******************************************************************************
-HTTP_BASE_URL="http://{{ include "hono.fullname" . }}-service-device-registry:8080/v1"
+URL_SCHEME="{{- if ( eq .Values.deviceRegistryExample.hono.registry.http.insecurePortEnabled true ) }}http{{ else }}https{{ end }}"
+URL_PORT=$([ "${URL_SCHEME}" = "http" ] && echo "8080" || echo "8443")
+HTTP_BASE_URL="${URL_SCHEME}://{{ include "hono.fullname" . }}-service-device-registry:${URL_PORT}/v1"
 
 check_status() {
   EXIT_STATUS=$1
@@ -35,7 +37,7 @@ add_tenant(){
   HTTP_REQUEST_BODY=$2
 
   echo "Adding tenant [$TENANT_ID]"
-  HTTP_RESPONSE=$(curl -o /dev/null -sw "%{http_code}" \
+  HTTP_RESPONSE=$(curl -o /dev/null -sw "%{http_code}" -k \
                     -X POST "$HTTP_BASE_URL/tenants/$TENANT_ID" \
                     --header 'Content-Type: application/json' \
                     --data-raw "$HTTP_REQUEST_BODY")
@@ -49,7 +51,7 @@ register_device(){
   HTTP_REQUEST_BODY=$3
 
   echo "Registering device [$TENANT_ID:$DEVICE_ID]"
-  HTTP_RESPONSE=$(curl -o /dev/null -sw "%{http_code}" \
+  HTTP_RESPONSE=$(curl -o /dev/null -sw "%{http_code}" -k \
                   -X POST "$HTTP_BASE_URL/devices/$TENANT_ID/$DEVICE_ID" \
                   --header 'Content-Type: application/json' \
                   --data-raw "$HTTP_REQUEST_BODY")
@@ -64,7 +66,7 @@ add_credentials(){
   HTTP_REQUEST_BODY=$3
 
   echo "Adding credentials [$TENANT_ID:$DEVICE_ID]"
-  HTTP_RESPONSE=$(curl -o /dev/null -sw "%{http_code}" \
+  HTTP_RESPONSE=$(curl -o /dev/null -sw "%{http_code}" -k \
                 -X PUT "$HTTP_BASE_URL/credentials/$TENANT_ID/$DEVICE_ID" \
                 --header 'Content-Type: application/json' \
                 --data-raw "$HTTP_REQUEST_BODY")
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -15,7 +15,7 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 2.5.5
+version: 2.5.6
 # Version of Hono being deployed by the chart
 appVersion: 2.4.0
 keywords:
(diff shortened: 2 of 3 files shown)


charts/hono/templates/dispatch-router/dispatch-router-example-keys.yaml
```diff
@@ -19,6 +19,7 @@ metadata:
   {{- include "hono.metadata" $args | nindent 2 }}
 type: Opaque
 data:
+  ca.crt: {{ .Files.Get "example/certs/ca-cert.pem" | b64enc }}
```
calohmn, line 22: Do we explicitly want to use `ca-cert.pem` and not `trusted-certs.pem` (which contains also `root-cert.pem` and is used in [example-trust-store.yaml](https://github.com/eclipse/packages/blob/bf3a4fd5e938c03f017e7526f64d7e857351e1b9/charts/hono/templates/defaults/example-trust-store.yaml#L20)) here?
(When running `mosquitto_sub` for example, I have to use the `trusted-certs.pem` for its `--cafile` param (e.g. in [setCloud2EdgeEnv.sh](https://github.com/eclipse/packages/pull/513/files#diff-51978e073eef3d7d78482c373ca55c804d6475aa70f2e37d68852c719d1c22acR99)), not `ca-cert.pem`.)

sophokles73: You're right, we should use the `trusted-certs.pem` instead.

calohmn: In [kafka-example-keys.yaml](https://github.com/eclipse/packages/blob/bf3a4fd5e938c03f017e7526f64d7e857351e1b9/charts/hono/templates/kafka/kafka-example-keys.yaml#L22) there is also the `ca-cert.pem` being used currently. Should we change that as well? The Secret is being set in the [existingSecrets](https://github.com/bitnami/charts/blob/dfc0dde55133111fcae8f27d83e227758e59bfe2/bitnami/kafka/values.yaml#L315) in the Kafka chart.

sophokles73: Hmmm, let's take a step back and reconsider this.
The example keys/certs are based on a (self-signed) root certificate, a CA certificate that has been signed using the root key and certificates for each of Hono's components which have been signed using the CA key.
So, in general, it should actually be sufficient for all of Hono's components that need to talk to each other to use the CA cert as their trust store because all other server certs have been signed using the CA key.
When using mosquitto_sub for connecting to the MQTT adapter it should also be sufficient to use the CA cert as the trust store because the MQTT adapter's server cert has been signed by the CA key as well.

sophokles73: Ok, I have taken another look. Actually, there is no need to add the trust store to the example keys secret because we already mount the generic hono-example-trust-store configmap into all containers, which contains the ca.crt (based on trusted-certs.pem).
I have therefore removed the change to the dispatch-router-example-keys ...

sophokles73:
Ok, I have taken another look. Actually, there is no need to add the trust store to the example keys secret because we already mount the generic hono-example-trust-store configmap into all containers, which contains the ca.crt (based on trusted-certs.pem).
I have therefore removed the change to the dispatch-router-example-keys ...

sophokles73:
@calohmn would you mind taking another look?

charts/hono/templates/dispatch-router/dispatch-router-example-keys.yaml
calohmn: > When using mosquitto_sub for connecting to the MQTT adapter it should also be sufficient to use the CA cert as the trust store because the MQTT adapter's server cert has been signed by the CA key as well.

In practise, this doesn't work.
When having installed the Hono chart in the default configuration and running
```
export MQTT_ADAPTER_IP=$(kubectl get service eclipse-hono-adapter-mqtt --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
cd <iot-packages-dir>
export MOSQUITTO_OPTIONS="--cafile $(pwd)/charts/hono/example/certs/ca-cert.pem --insecure"
mosquitto_sub -d -v -h ${MQTT_ADAPTER_IP} -p 8883 -u sensor1@DEFAULT_TENANT -P hono-secret ${MOSQUITTO_OPTIONS} -t command///req/#
```
there is this error:
```
Client null sending CONNECT
OpenSSL Error[0]: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed
Error: Protocol error
```
When using
`export MOSQUITTO_OPTIONS="--cafile $(pwd)/charts/hono/example/certs/root-cert.pem --insecure"`
or
`export MOSQUITTO_OPTIONS="--cafile $(pwd)/charts/hono/example/certs/trusted-certs.pem --insecure"`
instead, it works:
```
Client null sending CONNECT
Client null received CONNACK (0)
Client null sending SUBSCRIBE (Mid: 1, Topic: command///req/#, QoS: 0, Options: 0x00)
Client null received SUBACK
Subscribed (mid: 1): 0
```

See also:
```
$ openssl verify -verbose -CAfile root-cert.pem ca-cert.pem
ca-cert.pem: OK

$ openssl verify -verbose -CAfile ca-cert.pem mqtt-adapter-cert.pem
C = CA, L = Ottawa, O = Eclipse IoT, OU = Hono, CN = ca
error 2 at 1 depth lookup: unable to get issuer certificate
error mqtt-adapter-cert.pem: verification failed

$ openssl verify -verbose -CAfile root-cert.pem mqtt-adapter-cert.pem
C = CA, L = Ottawa, O = Eclipse IoT, OU = Hono, CN = mqtt-adapter
error 20 at 0 depth lookup: unable to get local issuer certificate
error mqtt-adapter-cert.pem: verification failed

$ openssl verify -verbose -CAfile <(cat ca-cert.pem root-cert.pem) mqtt-adapter-cert.pem
mqtt-adapter-cert.pem: OK
```

charts/hono/example/add_example_data_device_registry.sh
```diff
@@ -11,7 +11,11 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #*******************************************************************************
+{{- if ( eq .Values.deviceRegistryExample.hono.registry.http.insecurePortEnabled true ) }}
 HTTP_BASE_URL="http://{{ include "hono.fullname" . }}-service-device-registry:8080/v1"
+{{- else }}
+HTTP_BASE_URL="https://{{ include "hono.fullname" . }}-service-device-registry:8443/v1"
+{{- end }}
```
calohmn, line 18: Minor:
The helm template directive causes a bash syntax error to be displayed in the editor (preventing also shellcheck checks of the rest of the script).
This could be avoided by using something like:
```bash
URL_SCHEME="{{- if ( eq .Values.deviceRegistryExample.hono.registry.http.insecurePortEnabled true ) }}http{{ else }}https{{ end }}"
URL_PORT=$([ "$URL_SCHEME" = "http" ] && echo "8080" || echo "8443")
HTTP_BASE_URL="${URL_SCHEME}://{{ include "hono.fullname" . }}-service-device-registry:${URL_PORT}/v1"
```

calohmn:
> I have therefore removed the change to the dispatch-router-example-keys ...

The change is still mentioned in the commit message.

sophokles73:
@calohmn fixed the commit message

PR #516, opened 2023-10-28
Use Ditto 3.4.0, Hono 2.5.6 in c2e chart.

calohmn:
Also apply minor fix in post-install.sh, using the appropriate protocol for registering the default
tenant and device (adopted from 6cf96a989a083ab94461ffe6171be4e5cffd1bf2).

### DIFF
--- a/packages/cloud2edge/Chart.yaml
+++ b/packages/cloud2edge/Chart.yaml
@@ -10,9 +10,9 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #
-apiVersion: v1
-version: 0.7.0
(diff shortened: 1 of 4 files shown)


thjaeckle:
Ditto 3.4.0 requires a migration step.
And is not possible to do as rolling update.
So I would at least increase the minor version of the package and maybe we should start providing release notes..

calohmn:
@thjaeckle I've set the version to 0.8.0 now and updated the release notes entry with a note about the necessary migration.

thjaeckle (approved):
Cool, LGTM

PR #518, opened 2023-11-06
Allow other database than mysql

flobz:
This allow to specify a full springDatasourceUrl instead of create url from springDatasourceHost and springDatasourceDb.

### DIFF
--- a/charts/hawkbit/templates/deployment.yaml
+++ b/charts/hawkbit/templates/deployment.yaml
@@ -39,7 +39,7 @@ spec:
             - name: "SPRING_DATASOURCE_URL"
               {{- if .Values.env.springDatasourceUrl }}
               value: "{{ .Values.env.springDatasourceUrl }}"
-              {{ else }}
+              {{- else }}
               value: "jdbc:mysql://{{ if .Values.mysql.enabled }}{{ .Release.Name }}-mysql{{ else }}{{ .Values.env.springDatasourceHost }}{{ end }}:3306/{{ .Values.env.springDatasourceDb }}"
               {{- end }}
             - name: "SPRING_APPLICATION_JSON"
--- a/charts/hawkbit/templates/deployment.yaml
+++ b/charts/hawkbit/templates/deployment.yaml
@@ -37,7 +37,7 @@ spec:
             - name: SPRING_PROFILES_ACTIVE
               value: "{{ .Values.spring.profiles }}"
             - name: "SPRING_DATASOURCE_URL"
-              value: "jdbc:mysql://{{ if .Values.mysql.enabled }}{{ .Release.Name }}-mysql{{ else }}{{ .Values.env.springDatasourceHost }}{{ end }}:3306/{{ .Values.env.springDatasourceDb }}"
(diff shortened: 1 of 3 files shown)


flobz:
I updated chart version, workflow should pass.

charts/hawkbit/templates/deployment.yaml
```diff
@@ -37,7 +37,7 @@ spec:
             - name: SPRING_PROFILES_ACTIVE
               value: "{{ .Values.spring.profiles }}"
             - name: "SPRING_DATASOURCE_URL"
-              value: "jdbc:mysql://{{ if .Values.mysql.enabled }}{{ .Release.Name }}-mysql{{ else }}{{ .Values.env.springDatasourceHost }}{{ end }}:3306/{{ .Values.env.springDatasourceDb }}"
+              value: "{{ if .Values.env.springDatasourceUrl }}{{ .Values.env.springDatasourceUrl }}{{ else }}jdbc:mysql://{{ if .Values.mysql.enabled }}{{ .Release.Name }}-mysql{{ else }}{{ .Values.env.springDatasourceHost }}{{ end }}:3306/{{ .Values.env.springDatasourceDb }} {{end}}"
```
calohmn, line 40: An extra space character got added here in the last part of the string:
`[...].springDatasourceDb }} {{end}}"`.

In general, this long line with the 2 `if` conditions is quite hard to read.
How about making multiple lines out of it?
E.g.
```
              {{- if .Values.env.springDatasourceUrl }}
              value: "{{ .Values.env.springDatasourceUrl }}"
              {{- else }}
              value: "jdbc:mysql://{{ if .Values.mysql.enabled }}{{ .Release.Name }}-mysql{{ else }}{{ .Values.env.springDatasourceHost }}{{ end }}:3306/{{ .Values.env.springDatasourceDb }}"
              {{- end}}
```

charts/hawkbit/values.yaml
```diff
@@ -83,6 +83,7 @@ fileStorage:
 env:
   springDatasourceHost: "hawkbit-mysql"
   springDatasourceDb: "hawkbit"
+  springDatasourceUrl: ""
```
calohmn, line 86: I think a comment would be helpful here, pointing out that if `springDatasourceUrl` is set, it takes precedence over the URL created from default mysql DB host (if `mysql.enabled`) or `springDatasourceHost` and `springDatasourceDb`.

charts/hawkbit/templates/deployment.yaml
flobz: done

calohmn:
@strailov Any comments on this?

strailov:
Nope, looks good to me.

charts/hawkbit/templates/deployment.yaml
```diff
@@ -37,7 +37,11 @@ spec:
             - name: SPRING_PROFILES_ACTIVE
               value: "{{ .Values.spring.profiles }}"
             - name: "SPRING_DATASOURCE_URL"
+              {{- if .Values.env.springDatasourceUrl }}
+              value: "{{ .Values.env.springDatasourceUrl }}"
+              {{ else }}
```
avgustinmm, line 42:
```suggestion
              {{- else }}
```
if you have set env.springDatasourceUrl then in the result deployment file you will have an empty line after "value: ...". 
better to change it to "{{- else }}" in order to remove the excessive empty line

flobz:
@avgustinmm I apply your suggestion, can we merge ?

avgustinmm:
I'm not a committer in the project. 
I haven't checked the PR extensively - just commented about something I've seen on first look.

calohmn (approved):
LGTM, thanks for contributing.

PR #523, opened 2024-01-28
[#522] Update to Hono 2.5.0

### LINKED ISSUE #522: Update to Hono 2.5.0
Update Hono chart to use Hono 2.5.0. In doing so, we should also change the Apache Kafka broker configuration to use KRaft mode (i.e. without running Apache Zookeeper).

sophokles73:
Changed configuration to run Apache Kafka in KRaft mode.

Fixes #522

### DIFF
--- a/charts/hono/Chart.yaml
+++ b/charts/hono/Chart.yaml
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2019, 2023 Contributors to the Eclipse Foundation
+# Copyright (c) 2019 Contributors to the Eclipse Foundation
 #
 # See the NOTICE file(s) distributed with this work for additional
 # information regarding copyright ownership.
@@ -15,9 +15,9 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 2.5.6
+version: 2.6.0
 # Version of Hono being deployed by the chart
-appVersion: 2.4.0
+appVersion: 2.5.0
 keywords:
 - iot-chart
 - IoT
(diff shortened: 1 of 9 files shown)


sophokles73:
@calohmn Would you mind taking a look?

charts/hono/Chart.yaml
```diff
@@ -15,9 +15,9 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 2.5.6
+version: 2.5.7
```
calohmn, line 18: I think the changes in this PR would rather require a `2.6.0` version here.

charts/hono/values.yaml
```diff
@@ -1724,82 +1724,68 @@ kafka:
...
-  replicaCount: 1
   # Do not write data to a PersistentVolume by default.
   # When setting this to true, make sure to also set "kafka.zookeeper.persistence.enabled" to true and
   # make sure that Kafka is stopped cleanly using SIGTERM _before_ Zookeeper is stopped.
   # Otherwise you will most likely run into issues like
   # https://stackoverflow.com/questions/39759071/error-while-starting-kafka-broker
-  persistence:
-    enabled: false
+#  persistence:
+#    enabled: false
```
calohmn, line 1765: It looks like there is no top-level `persistence` property anymore in the Kafka-chart. 
Also the mentioning of the `kafka.zookeeper.persistence.enabled` setting above could be rather confusing, now that KRaft mode is used by default.

calohmn:
I think it would be good to also add support for the new `useLegacyTraceContextFormat` property in the new chart version.
I've created #524 for that (to be merged in this PR branch).

sophokles73:
sounds good, I will do that 👍

sophokles73:
@calohmn would you mind taking another look?

calohmn:
The Kafka chart update allows us to update the MongoDB chart again - see #490.
I think we could at least use the MongoDB chart version `^13.x` again (undoing the change from #491).

sophokles73:
@calohmn I have changed the MongoDB dependency ...

sophokles73:
@calohmn I will now squash this into two commits and then we're good to go, right?

calohmn (approved):
@sophokles73 yes, LGTM.

PR #524, opened 2024-01-30
[#522] Hono 2.5.0: Add useLegacyTraceContextFormat property in values.yaml.

calohmn:
This is for #522, to be included in PR #523:
Adds support for the `useLegacyTraceContextFormat` property introduced in Hono 2.5.0 (https://github.com/eclipse-hono/hono/issues/3597).

### DIFF
--- a/charts/hono/templates/hono-service-command-router/hono-service-command-router-secret.yaml
+++ b/charts/hono/templates/hono-service-command-router/hono-service-command-router-secret.yaml
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2021, 2022 Contributors to the Eclipse Foundation
+# Copyright (c) 2021 Contributors to the Eclipse Foundation
 #
 # See the NOTICE file(s) distributed with this work for additional
 # information regarding copyright ownership.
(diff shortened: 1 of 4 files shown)


charts/hono/templates/hono-service-command-router/hono-service-command-router-secret.yaml
```diff
@@ -88,6 +88,7 @@ stringData:
         certPath: {{ .Values.adapters.commandAndControlSpec.certPath }}
         trustStorePath: {{ .Values.adapters.commandAndControlSpec.trustStorePath }}
         hostnameVerificationRequired: {{ .Values.adapters.commandAndControlSpec.hostnameVerificationRequired }}
+        useLegacyTraceContextFormat: {{ .Values.adapters.commandAndControlSpec.useLegacyTraceContextFormat }}
```
sophokles73, line 91: Do we need/want to be able to set this property per component? IMHO it should be sufficient to set this at a global level, e.g. using something like `.Values.useLegacyTraceContextFormat`.

WDYT?

calohmn: I've added a top-level `useLegacyAmqpTraceContextFormat` property now.

calohmn:
@sophokles73 Could you have another look?

charts/hono/values.yaml
```diff
@@ -105,6 +105,14 @@ serviceType:
 messagingNetworkTypes:
 - "kafka"
 
+# useLegacyAmqpTraceContextFormat determines how OpenTelemetry trace context information
+# is stored in an AMQP 1.0 message sent to the AMQP Messaging Network. A "false" value here
+# configures the format used in previous Hono versions, while setting "true" chooses a more
+# generic format (compatible with e.g. Eclipse Ditto). Refer to
```
sophokles73, line 111: this sounds wrong, shouldn't it be the other way around, i.e.
`useLegacyFormat=true` means that I want to use the old (legacy) format, right?

calohmn: oh, yes. It's fixed now.

sophokles73 (approved):
LGTM

PR #538, opened 2024-06-03
Update to Hono 2.6.0

sophokles73:
Also removed @ctron as chart maintainer because he no longer works on Hono.

Addresses #537


charts/hono/Chart.yaml
```diff
@@ -15,9 +15,9 @@ name: hono
 description: |
   Eclipse Hono™ provides remote service interfaces for connecting large numbers of IoT devices to a back end and
   interacting with them in a uniform way regardless of the device communication protocol.
-version: 2.6.1
+version: 2.6.2
 # Version of Hono being deployed by the chart
-appVersion: 2.5.1
+appVersion: 2.6.0
```
calohmn, line 20: Changes corresponding to the renaming of the Command Router image (from `hono-service-command-router` to `hono-service-command-router-infinispan` are required as well, also concerning the native image).

Also the currently wrong `kafka.autoCreateTopicsEnable` and `kafka.deleteTopicEnable` properties in `values.yaml` should be replaced by adding 
```
      auto.create.topics.enable=true
      delete.topic.enable=true
```
to `kafka.controller.extraConfig`.

As a further point, the `profileOpenshift-values.yaml` could be corrected in terms of putting
```
  podSecurityContext:
    enabled: false
  containerSecurityContext:
    enabled: false
```
under `kafka.controller`.

sophokles73: > Changes corresponding to the renaming of the Command Router image (from hono-service-command-router to hono-service-command-router-infinispan are required as well, also concerning the native image).

The images have also been tagged with the original names. As long as there is no alternative (Redis based) implementation, there is little need for changing the names (yet).

sophokles73: Just for my understanding: the other issues are not related to (that is: caused by) the update to Hono 2.6.0, or am I mistaken?

calohmn: > The images have also been tagged with the original names. As long as there is no alternative (Redis based) implementation, there is little need for changing the names (yet).

Ok, I wasn't aware of that. I had tested the chart with locally built `2.6.0-SNAPSHOT` images (using the `honoImagesTag` helm install parameter) right before the 2.6.0 release and saw the need to adapt the Command Router image name with that (by setting `commandRouterService.imageName`). Maybe a corresponding hint can be added in the chart release notes.

> Just for my understanding: the other issues are not related to (that is: caused by) the update to Hono 2.6.0, or am I mistaken?

Yes, these are independent.

sophokles73:
@calohmn I have added the proposed changes. Thanks for spotting :+1:

sophokles73:
> I had tested the chart with locally built 2.6.0-SNAPSHOT images (using the honoImagesTag helm install parameter) right before the 2.6.0 release and saw the need to adapt the Command Router image name

You are raising a valid point. For local testing and deploying using the chart, it would indeed be helpful to use the new image names. I will change that ...

charts/hono/values.yaml
```diff
@@ -1457,7 +1457,7 @@ commandRouterService:
 
   # imageName contains the name (excluding registry)
   # of the container image to use for the Command Router service
-  imageName: "eclipse/hono-service-command-router"
+  imageName: "eclipse/hono-service-command-router-infinispan"
```
calohmn, line 1460: I think the other image references should be adapted as well:
- https://github.com/eclipse/packages/blob/044b6075f84139a00c3e95002ae74e8255598f06/charts/hono/ci/quarkus-native-images-values.yaml#L47
- https://github.com/eclipse/packages/blob/911c8fd57c22c86ecf7f77e67abf5d41e056c219/charts/hono/README.md?plain=1#L610

sophokles73:
@calohmn I have pushed corresponding changes ...

calohmn (approved):
LGTM

PR #544, opened 2024-06-26
[#543] Update Hono example certificates.

### LINKED ISSUE #543: Hono/Kafka Certificates Expired
It seems that the default certificates that Hono and Kafka are using have all expired 2 days ago.

I am getting certificate validation errors on all the hono adapter pods.

I've checked some of the certificates and they seem to only be valid until June 23, 2024.

would it be possible to get a new release with a refreshed set of certificates?

or a guide on how to replace all of them for a local deployment?

Thank you!

calohmn:
This fixes #543.


sophokles73 (approved):
LGTM

PR #546, opened 2024-06-27
[#543] Update Hono/Ditto versions in c2e chart.

calohmn:
Updates the cloud2edge chart with the new Hono chart, including the updated example certificates (#543). Ditto version is updated as well.


thjaeckle (approved):

sophokles73 (approved):
LGTM

PR #549, opened 2024-09-05
Update hawkBit to Application Version 0.5.0

lharzenetter:
In the current version, the helm chart does not support the latest version of hawkBit that is [0.5.0].

Because of the removal of the vaadin UI, the readyness and liveness probes are broken.
Additionally, by default, the version [0.5.0] uses the [MariaDB Connector][MariaDB-j] instead of the [MySQL Connector][mysql-j].

Thus, I updated the readiness and liveness probes to use the Swagger-UI and the MariaDB connection string by default.

  [0.5.0]: https://github.com/eclipse/hawkbit/releases/tag/0.5.0 
  [MariaDB-j]: https://mariadb.com/kb/en/about-mariadb-connector-j/
  [mysql-j]: https://mvnrepository.com/artifact/com.mysql/mysql-connector-j

### DIFF
--- a/charts/hawkbit/templates/tests/test-connection.yaml
+++ b/charts/hawkbit/templates/tests/test-connection.yaml
@@ -16,7 +16,7 @@ spec:
       command: ['curl']
       args: [
         "-X 'GET'",
-        "-u admin:admin",
+        "-u 'admin:admin'",
         "'http://{{ include "hawkbit.fullname" . }}:{{ .Values.service.port }}/rest/v1/userinfo'"
       ]
   restartPolicy: Never
--- a/charts/hawkbit/templates/tests/test-connection.yaml
+++ b/charts/hawkbit/templates/tests/test-connection.yaml
@@ -15,8 +15,8 @@ spec:
       image: curlimages/curl
       command: ['curl']
       args: [
-        "-X 'GET'",
-        "-u 'admin:admin'",
-        "'http://{{ include "hawkbit.fullname" . }}:{{ .Values.service.port }}/rest/v1/userinfo'"
+        "-X", "GET",
+        "-u", "admin:admin",
+        "http://{{ include "hawkbit.fullname" . }}:{{ .Values.service.port }}/rest/v1/userinfo"
       ]
   restartPolicy: Never
--- a/charts/hawkbit/templates/tests/test-connection.yaml
+++ b/charts/hawkbit/templates/tests/test-connection.yaml
@@ -16,7 +16,7 @@ spec:
       command: ['curl']
       args: [
         "-X", "GET",
-        "-u", "admin:admin",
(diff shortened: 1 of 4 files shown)


calohmn:
@lharzenetter The CI checks are currently failing here because of a failing `helm test [-n <NS>] eclipse-hawkbit` check.
The issue is that in `charts/hawkbit/templates/tests/test-connection.yaml` there is an HTTP request to the root resource on the hawkBit pod being made, which results in a status 400 response (see https://github.com/eclipse/hawkbit/issues/1832).
Using
```
args:  ['{{ include "hawkbit.fullname" . }}:{{ .Values.service.port }}/swagger-ui/index.html']
```
instead lets the test succeed.
However, as the same endpoint is already used in the liveness/readiness checks, this test wouldn't look very useful. Maybe another endpoint could be used (using the admin credentials).

@strailov @avgustinmm Could you have a look at this PR?

avgustinmm:
> However, as the same endpoint is already used in the liveness/readiness checks, this test wouldn't look very useful. Maybe another endpoint could be used (using the admin credentials).

Using admin credentials you could call:

```shell
curl -X 'GET' \
  'http://localhost:8080/rest/v1/userinfo' \
  -H 'accept: application/json' \
  -H 'Authorization: Basic ...'
```

This shall return info about the tenant and user, e.g.:

```json
{
  "tenant": "DEFAULT",
  "username": "admin"
}
```

This could be a useful liveness/readiness probe since it indicates that the management API is ready to be called up.

lharzenetter:
> This could be a useful liveness/readiness probe since it indicates that the management API is ready to be called up.

I agree.
However, IMHO, we should not use an admin resource as a liveness/readiness probe as you must update it if you use "real" credentials.

lharzenetter:
Do you have any comments on the code? Could you please merge it?

Cheers

strailov:
It looks good for me. Could be merged.

charts/hawkbit/templates/tests/test-connection.yaml
```diff
@@ -11,8 +11,12 @@ metadata:
...
-    - name: wget
-      image: busybox
-      command: ['wget']
-      args:  ['{{ include "hawkbit.fullname" . }}:{{ .Values.service.port }}']
+    - name: curl
+      image: curlimages/curl
+      command: ['curl']
+      args: [
+        "-X", "GET",
+        "-u", "admin:admin",
```
calohmn, line 19: I think instead of `admin` as username/password here, corresponding `Values` references should be put in here (excluding the `{noop}` password prefix).

charts/hawkbit/values.yaml
```diff
@@ -41,7 +41,7 @@ updateStrategy:
   type: Recreate
 
 nameOverride: ""
-fullnameOverride: ""
+fullnameOverride: "eclipse-hawkbit"
```
calohmn, line 44: Is this change needed? I think it's better to keep the old value and let the default be applied (including the release name).

And, could you do a rebase? There was an update of the chart version.

calohmn (approved):

PR #576, opened 2026-09-13
Upgrade to Jaeger 2.20.0

n-deliyski:


sophokles73:
@n-deliyski thanks for the PR
The chart also needs its (patch) version to be bumped.

n-deliyski:
The certificates valid one year have expired. I have regenerated them locally at my side during testing but excluded them from the PR.

sophokles73:
@n-deliysky without valid certificates, the deployment of the chart will fail. That is the reason why the checks fail ...
Can you please include the regenerated certs in the PR as a separate commit preceding the existing commit containing your changes?

sophokles73 (approved):
LGTM

sophokles73:
@calohmn any additional thoughts?

